PR #2, opened 2021-04-21
Fix issue with package resolution and version pinning.

bsneed:
- Version pinned Sovran-Swift to 1.0.0+.
- Fixed Package.resolved.
- Updated xcodeproj to cover the same details.


migs647 (approved):
LGTM

PR #9, opened 2021-05-12
Removes need for analytics param in Plugin init

bsneed:
- Removes need for analytics param in Plugin init
- Adds test to verify plugin's analytics property is set appropriately even when done out of order.


migs647 (approved):
LGTM ... for what it is.

PR #11, opened 2021-05-18
Better handle errors where conversion occurs.

bsneed:
- Adds exceptionFailure method.
- Better reports errors to developers where conversions or critical errors might occur.


Sources/Segment/Utilities/Utils.swift
```diff
@@ -0,0 +1,26 @@
...
+    let env = ProcessInfo().environment
+    let value = (env["XCTestConfigurationFilePath"] != nil)
+    return value
+}()
+
+internal let _SegmentException = "SegmentException"
+internal func exceptionFailure(_ message: String) {
+    let args: [CVarArg] = []
+    if isUnitTesting {
+        NSException.raise(NSExceptionName(_SegmentException), format: message, arguments: getVaList(args))
```
migs647, line 20: Are we sure we want to use NSExceptions? They aren't friendly with recovery, and apple doesn't recommend usage unless you fail / exit

bsneed: Yeah, that's why I used them.  I want a hard stop when unit testing or during development.

migs647: Awesome, if that is the intention then we're good :)

PR #14, opened 2021-05-25
Adding Examples of Apps, Plugins and Tasks.

bsneed:


Examples/apps/SegmentUIKitExample/SegmentUIKitExample/Tab1ViewController.swift
```diff
@@ -0,0 +1,24 @@
...
+
+class Tab1ViewController: UITableViewController {
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        // Do any additional setup after loading the view.
+    }
+
+    @IBAction func queryAction(_ sender: Any) {
+        let alertController = QueryAlertController(title: "I know you like pie", message: "Everyone likes pie", preferredStyle: .alert)
```
migs647, line 19: What is this for?

Examples/destination_plugins/FlurryDestination.swift
```diff
@@ -0,0 +1,137 @@
...
+import Segment
+/**
+ An implmentation of the Flurry Analytics device mode destination as a plugin.
+ 
+ NOTE: This is currently commented out due to lack of arm64 support in the flurry SDK via swift package manager.
+ See here: https://github.com/flurry/FlurrySwiftPackage/issues/3
+ */
+
+/*
+import FlurryAnalytics
```
migs647, line 48: This worked for me, we sure we need to comment it out? I just need to do a clean build each time on Xcode 12.5. 12.4 it works without having to do that /eyeroll

migs647 (approved):
Just a few questions -- but definitely good enough to merge.

Examples/destination_plugins/FlurryDestination.swift
bsneed: It works for you because you don't have an M1 Mac.  That's the issue.

Examples/apps/SegmentUIKitExample/SegmentUIKitExample/Tab1ViewController.swift
bsneed: it's an example of customized screen tracking.  just added a comment saying such.

Examples/destination_plugins/FlurryDestination.swift
migs647: Interesting, because I was getting that "warning" which resulted in a build error until then. But I'm with you. I think we're going to run into that on almost all of these.

PR #15, opened 2021-06-01
Add Adjust device mode integration sample & example app

bsneed:


alanjcharles (approved):

Examples/destination_plugins/AdjustDestination.swift
```diff
@@ -0,0 +1,210 @@
...
+    public func update(settings: Settings) {
+        guard let settings: AdjustSettings = settings.integrationSettings(name: "Adjust") else { return }
+        self.settings = settings
+        
+        var environment = ADJEnvironmentSandbox
+        if let _ = settings.setEnvironmentProduction {
+            environment = ADJEnvironmentProduction
+        }
+        
+        let adjustConfig = ADJConfig(appToken: settings.appToken, environment: environment)
```
migs647, line 76: Do we need to protect against an empty appToken? Can it be nil?

Examples/destination_plugins/AdjustDestination.swift
```diff
@@ -0,0 +1,210 @@
...
+            adjustConfig?.delegate = self
+        }
+        
+        if let useDelay = settings.setDelay, useDelay == true, let delayTime = settings.delayTime {
+            adjustConfig?.delayStart = delayTime
+        }
+        
+        Adjust.appDidLaunch(adjustConfig)
+
+        started = true
```
migs647, line 92: Do we want to queue up any of these events if started hasn't happened yet?

.. If so, should we consider adding that feature to the core library?

migs647 (approved):
LGTM

Examples/destination_plugins/AdjustDestination.swift
bsneed: It potentially could be, but let adjust report that error.

Examples/destination_plugins/AdjustDestination.swift
bsneed: Possibly; working on that separately from this PR.

PR #17, opened 2021-06-02
Fixes + Misses corrections

bsneed:
- Removed segment example scheme.
- Connected flush/reset such that all plugins can use.
- Fixed version number mess-up.
- Added remote notification and reset support to Adjust plugin.


migs647 (approved):
LGTM

PR #18, opened 2021-06-02
Added Mixpanel Destination

migs647:


Examples/destination_plugins/MixpanelDestination.swift
```diff
@@ -0,0 +1,302 @@
...
+            }
+            
+            mixpanel?.setGroup(groupKey: key, groupID: nsGroupID)
+            analytics?.log(message: "Mixpanel setGroup \(key) groupID \(groupID)")
+        }
+        
+        return event
+    }
+    
+    func alias(event: AliasEvent) -> AliasEvent? {
```
bsneed, line 173: we've explicitly disabled alias in analytics-swift, fyi.  not sure how critical this bit is to overall operation.

Examples/destination_plugins/MixpanelDestination.swift
```diff
@@ -0,0 +1,302 @@
...
+        analytics?.log(message: "Mixpanel reset")
+    }
+    
+    func flush() {
+        mixpanel?.flush()
+        analytics?.log(message: "Mixpanel Flush")
+    }
+    
+    // MARK: - Private Methods
+    private func mixpanelTrack(_ eventName: String, properties: [String: Any]? = nil) {
```
bsneed, line 202: might be good to move non-protocol conformance methods out into an extension.

bsneed (approved):

Examples/destination_plugins/MixpanelDestination.swift
migs647: Done

Examples/destination_plugins/MixpanelDestination.swift
migs647: Left for now... lets re-evaluate after discussions.

PR #24, opened 2021-06-04
Bsneed/queueing

bsneed:
- Added `started` property to System state.
- Added `StartupQueue` plugin that will effectively pause event processing until plugins have started.
- Accounts for plugins added much later and resumes queueing until started.


migs647 (approved):
LGrrrrreatTM 🐯

prayansh (approved):
🔥

migs647 (approved):
LGTM with limits

PR #25, opened 2021-06-09
Added Firebase destination

migs647:
To test, download and replace the GoogleService-Info.plist from your Firebase project.


bsneed (approved):
Looks really good!!

PR #26, opened 2021-06-10
Add StartupQueue automatically; update tests.

bsneed:
- Adds StartupQueue to the initial list of plugins.
- Adds waitUntilStarted(analytics:) to test suite.
- Modifies some tests to wait until started.

### DIFF
--- a/Sources/Segment/Settings.swift
+++ b/Sources/Segment/Settings.swift
@@ -23,7 +23,7 @@ public struct Settings: Codable {
     
     public init(from decoder: Decoder) throws {
         let values = try decoder.container(keyedBy: CodingKeys.self)
-        integrations = try? values.decode(JSON.self, forKey: CodingKeys.integrations)
+        self.integrations = try? values.decode(JSON.self, forKey: CodingKeys.integrations)
         self.plan = try? values.decode(JSON.self, forKey: CodingKeys.plan)
         self.edgeFunctions = try? values.decode(JSON.self, forKey: CodingKeys.edgeFunctions)
     }
@@ -52,18 +52,49 @@ public struct Settings: Codable {
 
 }
 
+extension Settings: Equatable {
+    public static func == (lhs: Settings, rhs: Settings) -> Bool {
+        let l = lhs.prettyPrint()
+        let r = rhs.prettyPrint()
+        return l == r
+    }
+}
+
 extension Analytics {
-    func checkSettings() {
+    internal func update(settings: Settings) {
+        apply { (plugin) in
+            // tell all top level plugins to update.
+            update(plugin: plugin, settings: settings)
+        }
+    }
+    
+    internal func update(plugin: Plugin, settings: Settings) {
+        plugin.update(settings: settings)
+        // if it's a destination, tell it's plugins to update as well.
+        if let dest = plugin as? DestinationPlugin {
+            dest.apply { (subPlugin) in
+                subPlugin.update(settings: settings)
+            }
+        }
+    }
+    
+    internal func checkSettings() {
         let writeKey = self.configuration.values.writeKey
         let httpClient = HTTPClient(analytics: self, cdnHost: configuration.values.cdnHost)
+        // stop things; queue in case our settings have changed.
+        self.store.dispatch(action: System.ToggleRunningAction(running: false))
         httpClient.settingsFor(writeKey: writeKey) { (success, settings) in
             if success {
                 if let s = settings {
                     // put the new settings in the state store.
                     // this will cause them to be cached.
                     self.store.dispatch(action: System.UpdateSettingsAction(settings: s))
+                    // let plugins know we just received some settings..
+                    self.update(settings: s)
                 }
             }
+            // we're good to go back to a running state.
+            self.store.dispatch(action: System.ToggleRunningAction(running: true))
         }
     }
 }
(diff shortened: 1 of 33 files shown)


prayansh (approved):

Tests/Segment-Tests/Support/TestUtilities.swift
```diff
@@ -107,3 +107,16 @@ class OutputReaderPlugin: Plugin {
...
+func waitUntilStarted(analytics: Analytics?) {
+    guard let analytics = analytics else { return }
+    var started = false
+    while started == false {
+        if let system: System = analytics.store.currentState() {
+            if system.started {
+                started = true
+            }
+        }
+        RunLoop.main.run(until: Date.distantPast)
```
migs647, line 120: Would like a comment here saying you're ending the current runloop incase some newbs don't recognize it and take it out.

Tests/Segment-Tests/Support/TestUtilities.swift
```diff
@@ -107,3 +107,16 @@ class OutputReaderPlugin: Plugin {
         return event
     }
 }
+
+func waitUntilStarted(analytics: Analytics?) {
+    guard let analytics = analytics else { return }
+    var started = false
+    while started == false {
+        if let system: System = analytics.store.currentState() {
```
migs647, line 115: Wouldn't mind some commented context =). Makes sense right now in the moment, but what about three months from now?

Tests/Segment-Tests/Support/TestUtilities.swift
migs647: nevermind, just noticed this is in tests

Tests/Segment-Tests/Support/TestUtilities.swift
migs647: nevermind, just noticed this is in tests

migs647 (approved):
LGTM

Sources/Segment/Utilities/QueueTimer.swift
```diff
@@ -0,0 +1,90 @@
...
+            return
+        }
+        state = .resumed
+        timer.resume()
+    }
+}
+
+
+extension TimeInterval {
+    static func milligarvins(_ value: Int) -> TimeInterval {
```
prayansh: i think u got a typo here 😂

migs647: 🤦‍♀️

Sources/Segment/Settings.swift
```diff
@@ -52,18 +52,49 @@ public struct Settings: Codable {
...
+        return l == r
+    }
+}
+
 extension Analytics {
-    func checkSettings() {
+    internal func update(settings: Settings) {
+        apply { (plugin) in
+            // tell all top level plugins to update.
+            update(plugin: plugin, settings: settings)
```
migs647, line 67: Do we want to maybe discriminate on only destination plugins here or do we want to track "started" on all types of plugins? I'm cool with it being all, just want to make sure we're on the same page.

Sources/Segment/Settings.swift
```diff
@@ -79,10 +79,23 @@ extension Analytics {
     }
     
     internal func checkSettings() {
+        if isUnitTesting {
+            // we don't really wanna wait for this network call during tests...
+            // but we should make it work similarly.
+            store.dispatch(action: System.ToggleRunningAction(running: false))
+            DispatchQueue.main.async {
```
migs647, line 86: Do we want `[weak self] in` for the closure?

Sources/Segment/Startup.swift
```diff
@@ -66,34 +66,46 @@ extension Analytics: Subscriber {
...
 extension Analytics {
     internal func setupSettingsCheck() {
+        // do the first one
+        checkSettings()
+        // set up return-from-background to do it again.
         NotificationCenter.default.addObserver(forName: UIApplication.willEnterForegroundNotification, object: nil, queue: OperationQueue.main) { (notification) in
-            self.checkSettings()
+            guard let app = notification.object as? UIApplication else { return }
+            if app.applicationState == .background {
+                self.checkSettings()
```
migs647, line 79: I'm still unsettled about using lifecycle events for triggering this.

Sources/Segment/Startup.swift
```diff
@@ -66,34 +66,46 @@ extension Analytics: Subscriber {
...
 import Cocoa
 extension Analytics {
     internal func setupSettingsCheck() {
-        NotificationCenter.default.addObserver(forName: NSApplication.willBecomeActiveNotification, object: nil, queue: OperationQueue.main) { (notification) in
+        // do the first one
+        checkSettings()
+        // now set up a timer to do it every 24 hrs.
+        // mac apps change focus a lot more than iOS apps, so this
+        // seems more appropriate here.
+        QueueTimer.schedule(interval: 5, queue: .main) {
```
migs647: This is so much better than using a selector 👍

Sources/Segment/Utilities/QueueTimer.swift
```diff
@@ -0,0 +1,90 @@
...
+        return TimeInterval(other: ((value / 60) * 1000) / 42)
```
migs647: Funny guy... but what is the 42 for?

migs647 (approved):
LGTM

Sources/Segment/Settings.swift
bsneed: yeah, should be all.

Sources/Segment/Settings.swift
bsneed: yes, good call out.

PR #31, opened 2021-07-12
Updated lifecycle events for AmplitudeSession

migs647:


bsneed:
it’s inserting its metadata even if amplitude isn’t in the list of destinations.  seems like it shouldn’t do that.  wdyt?

bsneed (approved):

PR #33, opened 2021-08-03
Add release automation script.

bsneed:
- Added release.sh
- Updated releasing.md
- Added safety text in Version.swift to deter by-hand modification.

### DIFF
--- a/Sources/Segment/Version.swift
+++ b/Sources/Segment/Version.swift
@@ -7,5 +7,12 @@
 
 // Referred to by Analytics.swift
 
+// DO NOT MODIFY THIS FILE BY HAND!!
+// DO NOT MODIFY THIS FILE BY HAND!!
+// DO NOT MODIFY THIS FILE BY HAND!!
+// DO NOT MODIFY THIS FILE BY HAND!!
+
+// Use release.sh's automation.
+
 // BREAKING.FEATURE.FIX
 internal let __segment_version = "1.0.1"
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,110 @@
+#!/bin/bash
+
+vercomp () {
+	if [[ $1 == $2 ]]
+	then
+		return 0
+	fi
+	local IFS=.
+	local i ver1=($1) ver2=($2)
+	# fill empty fields in ver1 with zeros
+	for ((i=${#ver1[@]}; i<${#ver2[@]}; i++))
+	do
+		ver1[i]=0
+	done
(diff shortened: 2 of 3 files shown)


Sources/Segment/Version.swift
```diff
@@ -7,5 +7,12 @@
 
 // Referred to by Analytics.swift
 
+// DO NOT MODIFY THIS FILE BY HAND!!
+// DO NOT MODIFY THIS FILE BY HAND!!
+// DO NOT MODIFY THIS FILE BY HAND!!
+// DO NOT MODIFY THIS FILE BY HAND!!
```
prayansh, line 13: 😂

release.sh
```diff
@@ -0,0 +1,110 @@
...
+# check that we're on the `main` branch
+branch=$(git rev-parse --abbrev-ref HEAD)
+if [ branch != 'main' ] 
+then
+	echo "The 'main' branch must be the current branch out to make a release."
+	echo "You are currently on: $branch"
+	exit 1
+fi
+
+versionFile="./sources/Segment/version.swift"
```
prayansh: not sure if `V` needs to be uppercase

release.sh
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
+
+vercomp () {
+	if [[ $1 == $2 ]]
+	then
+		return 0
```
prayansh, line 6: if the new version is the same as old version, we should error out, right?

release.sh
```diff
@@ -0,0 +1,110 @@
...
+	echo "New version must be greater than previous version ($version)."
+	exit 1
+fi
+
+# get the commits since the last release...
+# note: we do this here so the "Version x.x.x" commit doesn't show up in logs.
+changelog=git log --pretty=format:"- (%an) %s" $(git describe --tags --abbrev=0 @^)..@
+tempFile=$(mktemp)
+#write changelog to temp file.
+echo $changelog >> $tempFile
```
prayansh, line 94: whats the reason for a temp file?

prayansh (approved):
added some comments, but looks good

release.sh
bsneed: yup

release.sh
bsneed: case insensitive file system; but yeah it should be.

release.sh
bsneed: I write the changelog out there, and then gh release will read it in from there.

PR #35, opened 2021-08-04
Plugin Interface refactor

bsneed:
- Removes the `name` property entirely.
- Removes need for `plugin.init(name: )`
- Changes `analytics.find(pluginName: )` to be `analytics.find(pluginType: )`
- Adds `key` property on DestinationPlugin to be used for settings lookup and integration name.
- Makes all necessary example app/plugin changes.

### DIFF
--- a/Sources/Segment/Settings.swift
+++ b/Sources/Segment/Settings.swift
@@ -59,11 +59,13 @@ public struct Settings: Codable {
         return integrationSettings(forKey: plugin.key)
     }
     
-    /*
-    public func isDestinationEnabled(name: String) -> Bool {
+    public func isDestinationEnabled(key: String) -> Bool {
+        guard let settings = integrations?.dictionaryValue else { return false }
+        if settings.keys.contains(key) {
+            return true
+        }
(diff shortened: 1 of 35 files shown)


Sources/Segment/Settings.swift
```diff
@@ -40,20 +40,30 @@ public struct Settings: Codable {
...
         var result: T? = nil
         guard let settings = integrations?.dictionaryValue else { return nil }
-        if let dict = settings[name], let jsonData = try? JSONSerialization.data(withJSONObject: dict) {
+        if let dict = settings[key], let jsonData = try? JSONSerialization.data(withJSONObject: dict) {
             result = try? JSONDecoder().decode(T.self, from: jsonData)
         }
         return result
     }
+    
+    public func integrationSettings<T: Codable>(forPlugin plugin: DestinationPlugin) -> T? {
```
prayansh, line 58: o.O i like this 💯

prayansh (approved):
Looks good to me, but i aint a swift expert

PR #36, opened 2021-08-05
Modify destinations to avoid processing if they don't have settings

bsneed:
- Modified timeline processing such that we skip processing destinations that don't have settings.
- Added some convenience stuff necessary for tests.
- Added tests.

### DIFF
--- a/Sources/Segment/Settings.swift
+++ b/Sources/Segment/Settings.swift
@@ -21,6 +21,15 @@ public struct Settings: Codable {
         ])
     }
     
+    public init(writeKey: String) {
+        integrations = try! JSON([
+            SegmentDestination.Constants.integrationName.rawValue: [
+                SegmentDestination.Constants.apiKey.rawValue: writeKey,
+                SegmentDestination.Constants.apiHost.rawValue: HTTPClient.getDefaultAPIHost()
+            ]
+        ])
+    }
+    
     public init(from decoder: Decoder) throws {
         let values = try decoder.container(keyedBy: CodingKeys.self)
         self.integrations = try? values.decode(JSON.self, forKey: CodingKeys.integrations)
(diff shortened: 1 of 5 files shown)


Sources/Segment/Settings.swift
```diff
@@ -21,6 +21,15 @@ public struct Settings: Codable {
         ])
     }
     
+    public init(writeKey: String) {
```
prayansh, line 24: ohh i might have missed this on kotlin 😅

Sources/Segment/Timeline.swift
```diff
@@ -209,37 +209,40 @@ extension DestinationPlugin {
...
-            case let e as ScreenEvent:
-                destinationResult = screen(event: e) as? E
-            case let e as GroupEvent:
-                destinationResult = group(event: e) as? E
-            case let e as AliasEvent:
-                destinationResult = alias(event: e) as? E
-            default:
-                break
+
+        if let enabled = analytics?.settings()?.isDestinationEnabled(key: self.key), enabled == true {
```
prayansh, line 218: 👍

Sources/Segment/Timeline.swift
```diff
@@ -209,37 +209,40 @@ extension DestinationPlugin {
...
+            
+            // apply .after plugins ...
+            result = timeline.applyPlugins(type: .after, event: destinationResult)
         }
         
-        // apply .after plugins ...
-        let afterResult = timeline.applyPlugins(type: .after, event: destinationResult)
-
-        return afterResult
+        return result
```
prayansh, line 245: so does it return `null` or `incomingEvent` if the destination is disabled?

prayansh (approved):
LGTM. just a question about the destination processing return value

Sources/Segment/Timeline.swift
bsneed: it returns incomingEvent.  it ultimately ends up going into the ether, but it is consistent across the board.

PR #37, opened 2021-08-06
Fixes issue w/ UIApplication.shared not being supported in iOS App Extensions

### LINKED ISSUE #34: Xcode 13b4: 'shared' is unavailable in application extensions for iOS: Use view controller based solutions where appropriate instead.
```
…/SourcePackages/checkouts/analytics-swift/Sources/Segment/Plugins/Platforms/iOS/iOSLifecycleMonitor.swift:58:37: 'shared' is unavailable in application extensions for iOS: Use view controller based solutions where appropriate instead.
```

Xcode 13b4 introduced (erroneously, I think) a bug in which restrictions on APIs like being unavailable in extensions trigger errors, even when building applications. The usual fix for this is to declare the same restriction on the library's published methods.

You get this error by trying to build an iOS app in Xcode 13b4 with Segment as an SPM dependency.

bsneed:
- Fixes #34 
- Adds isAppExtension check.
- Changes application values in lifecycle func's to be optional.
- Generally accounts for the possibility that we may not have a UIApplication to pass around.

### DIFF
--- a/Sources/Segment/Plugins/Platforms/iOS/iOSLifecycleMonitor.swift
+++ b/Sources/Segment/Plugins/Platforms/iOS/iOSLifecycleMonitor.swift
@@ -203,7 +203,10 @@ extension SegmentDestination.UploadTaskInfo {
 
 extension UIApplication {
     static var safeShared: UIApplication? {
-        // UIApplication.shared is not available in app extensions.
+        // UIApplication.shared is not available in app extensions so try to get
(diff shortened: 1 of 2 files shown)


alanjcharles (approved):
looks good to me!

migs647 (approved):
LGTM. Wouldn't mind some comment cleanup on the safeShared accessor. Was a bit confusing at first.

PR #38, opened 2021-08-12
Adds key mapping support for use in destination plugins.

bsneed:
- Adds `mapKeys` function to JSON object w/ optional value transform step.
- Adds tests for key mapping functionality.

### DIFF
--- a/Sources/Segment/Utilities/JSON.swift
+++ b/Sources/Segment/Utilities/JSON.swift
@@ -269,6 +269,21 @@ extension JSON {
 
 // MARK: - Mapping
 
+extension JSON {
+    /// Maps keys supplied, in the format of ["Old": "New"].  Gives an optional value transformer that can be used to transform values based on the final key name.
+    /// - Parameters:
+    ///   - keys: A dictionary containing key mappings, in the format of ["Old": "New"].
+    ///   - valueTransform: An optional value transform closure.  Key represents the new key name.
+    public func mapKeys(_ keys: [String: String], valueTransform: ((_ key: String, _ value: Any) -> Any)? = nil) throws -> JSON {
+        guard let dict = self.dictionaryValue else { return self }
+        let mapped = try dict.mapKeys(keys, valueTransform: valueTransform)
+        let result = try JSON(mapped)
+        return result
+    }
+}
+
+// MARK: - Helpers
+
 extension Dictionary where Key == String, Value == Any {
     internal func mapKeys(_ keys: [String: String], valueTransform: ((_ key: Key, _ value: Value) -> Any)? = nil) throws -> [Key: Value] {
         let mapped = Dictionary(uniqueKeysWithValues: self.map { key, value -> (Key, Value) in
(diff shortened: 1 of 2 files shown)


Sources/Segment/Utilities/JSON.swift
```diff
@@ -267,9 +267,73 @@ extension JSON {
...
+        let mapped = try dict.mapKeys(keys, valueTransform: valueTransform)
+        let result = try JSON(mapped)
+        return result
+    }
+}
 
 // MARK: - Helpers
 
+extension Dictionary where Key == String, Value == Any {
+    internal func mapKeys(_ keys: [String: String], valueTransform: ((_ key: Key, _ value: Value) -> Any)? = nil) throws -> [Key: Value] {
```
migs647, line 288: Do you need `= nil` on there?

Sources/Segment/Utilities/JSON.swift
```diff
@@ -267,9 +267,73 @@ extension JSON {
...
+            if keys.keys.contains(key) {
+                if let mappedKey = keys[key] {
+                    // if so, lets change the key to the new value.
+                    newKey = mappedKey
+                }
+            }
+            // is this value a dictionary?
+            if let dictValue = value as? [Key: Value] {
+                if let r = try? dictValue.mapKeys(keys, valueTransform: valueTransform) {
+                    // if so, lets recurse...
```
migs647, line 303: 👍🏻

Sources/Segment/Utilities/JSON.swift
```diff
@@ -267,9 +267,73 @@ extension JSON {
...
+                    newValue = r
+                }
+            } else if let arrayValue = value as? [Value] {
+                // if it's an array, we need to see if any dictionaries are within and process
```
migs647, line 307: 👍🏻

Sources/Segment/Utilities/JSON.swift
```diff
@@ -267,9 +267,73 @@ extension JSON {
...
+                        }
+                    }
+                    return newValue
+                }
+            }
+            
+            if !(newValue is [Key: Value]), let transform = valueTransform {
+                // it's not a dictionary apply our transform.
+                
+                // note: if it's an array, we've processed any dictionaries inside
```
migs647, line 323: 👍🏻

Tests/Segment-Tests/JSON_Tests.swift
```diff
@@ -130,4 +130,51 @@ class JSONTests: XCTestCase {
...
+            if let v = newValue as? Int {
+                if v == 1 {
+                    newValue = 11
+                }
+            }
+            print("value = \(value.self)")
+            return newValue
+        }).dictionaryValue
+        
+        XCTAssertTrue(output!["AKey1"] as! Int == 11)
```
migs647, line 167: Can we change these to as? so we fail the unit test instead of crash the unit tests?

migs647 (approved):
LGTM -- just some nitpicks

alanjcharles (approved):
looks great. Awesome idea.

Sources/Segment/Utilities/JSON.swift
bsneed: Yeah, cuz I want it to be optional.  They likely just want to map keys, at which point, the values won't be transformed.

Tests/Segment-Tests/JSON_Tests.swift
bsneed: We could I guess, though that doesn't really bother me.  We'd need to go hit a bunch of other tests as well as that's my MO for unit tests.  FORCE UNWRAP ALL THE THINGS!

PR #44, opened 2021-08-19
Fix context issue and macOS vendor system miss.

### LINKED ISSUE #41: Setting trackApplicationLifecycleEvents to false strips context
Hi,

I've noticed that if `trackApplicationLifecycleEvents` is set to false, the context is stripped from all events.

I imagine this line should sit before the conditional block:

https://github.com/segmentio/analytics-swift/blob/ebdd751f1a04030c78ff4bdc787e0519f6bfe21f/Sources/Segment/Startup.swift#L44

One other thing: I would actually like to use application lifecycle events but I prefer a different naming scheme. Is there someway to set this manually? Thanks!

bsneed:
- Moves the addition of the context plugin OUTSIDE of the lifecycle check. Fixes #41 
- Adds a check to return the macOS vendor system instead of the empty default implementation.
- Adds test to verify context is present with the correct top level keys.


migs647 (approved):
LGTM

PR #45, opened 2021-08-19
Improve startup plugin collection.

bsneed:
- Moved required per-platform plugins to VendorSystem.
- Refactored startup sequence to fetch from VendorSystem and then apply lifecycle stuff.

### DIFF
--- a/Sources/Segment/Startup.swift
+++ b/Sources/Segment/Startup.swift
@@ -40,20 +40,24 @@ extension Analytics: Subscriber {
         // add context plugin as well as it's platform specific internally.
         // this must come first.
         plugins.append(Context())
+        
+        plugins += VendorSystem.current.requiredPlugins
 
         // setup lifecycle if desired
         if configuration.values.trackApplicationLifecycleEvents {
             #if os(iOS) || os(tvOS)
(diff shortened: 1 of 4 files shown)


Sources/Segment/Startup.swift
```diff
@@ -40,20 +40,24 @@ extension Analytics: Subscriber {
...
         // this must come first.
         plugins.append(Context())
+        
+        plugins += VendorSystem.current.requiredPlugins
 
         // setup lifecycle if desired
         if configuration.values.trackApplicationLifecycleEvents {
             #if os(iOS) || os(tvOS)
-            plugins += [iOSLifecycleMonitor(), iOSLifecycleEvents(), DeviceToken()]
+            plugins.append(iOSLifecycleEvents())
```
prayansh, line 49: would it make more sense to provide the lifecycleEvents plugin via the vendor system as well?

bsneed: I think that would add clutter in the long term if/when we have more things that fall into that bucket.

prayansh (approved):
LGTM, just a suggestion on providing the lifecycleEvent plugin via vendor system, to potentially remove macro code

migs647 (approved):
LGTM

PR #46, opened 2021-08-20
Added Logger and SegmentLogger plugin.

migs647:
Added Logger and SegmentLogger plugin. Added print target and system target (osLog). Contains flag for sending based on if settings is updated to stop logging.


Examples/other_plugins/UIKitScreenTracking.swift
```diff
@@ -122,7 +122,7 @@ extension UIViewController {
         guard let top = Self.seg__visibleViewController(activeController()) else { return }
         
         var name = String(describing: top.self.classForCoder).replacingOccurrences(of: "ViewController", with: "")
-        print(name)
+        analytics?.log(message: name)
```
bsneed: this should probably be more descriptive than just the name of the view.  ie: "Auto-tracking Screen: \(name)"

Examples/other_plugins/ConsoleLogger.swift
```diff
@@ -53,17 +53,17 @@ class ConsoleLogger: Plugin {
...
         }
         return event
     }
     
     // we also want to know when settings are retrieved or changed.
     func update(settings: Settings) {
         let json = settings.prettyPrint()
-        print("settings updated on instance: \(name)")
-        print("\(json)\n")
+        analytics?.log(message: "settings updated on instance: \(name)")
```
bsneed: these can probably be combined into a single call.

Examples/destination_plugins/AppsFlyerDestination.swift
```diff
@@ -250,7 +250,7 @@ extension AppsFlyerDestination: AppsFlyerLibDelegate {
 extension AppsFlyerDestination: DeepLinkDelegate, UIApplicationDelegate {
     
     func didResolveDeepLink(_ result: DeepLinkResult) {
-        print(result)
+        analytics?.log(message: result)
```
bsneed: this should be more descriptive rather than just the deep link result.

bsneed (approved):

bsneed (approved):

PR #50, opened 2021-08-23
Fixed Appsflyer plugin conformance issues.

bsneed:


alanjcharles (approved):
That...makes a lot more sense.

PR #59, opened 2021-09-09
Reduce file descriptor consumption

bsneed:
Reduces file descriptor usage by storing file handle rather than opening a new one with every write.

### DIFF
--- a/Sources/Segment/Utilities/Storage.swift
+++ b/Sources/Segment/Utilities/Storage.swift
@@ -283,7 +283,9 @@ extension Storage {
             if let endData = endData {
                 fileHandle?.seekToEndOfFile()
                 fileHandle?.write(endData)
-                try? fileHandle?.synchronize()
+                if #available(tvOS 13, *) {
+                    try? fileHandle?.synchronize()
+                }
                 fileHandle?.closeFile()
                 fileHandle = nil
             } else {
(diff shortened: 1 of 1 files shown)


Sources/Segment/Utilities/Storage.swift
```diff
@@ -241,23 +243,20 @@ extension Storage {
...
-                    handle.write(jsonData)
-                    handle.closeFile()
-                } else {
-                    assert(false, "Storage: Unable to convert event to json!")
+            let jsonString = event.toString()
+            if let jsonData = jsonString.data(using: .utf8) {
+                fileHandle?.seekToEndOfFile()
+                // prepare for the next entry
+                if newFile == false {
+                    fileHandle?.write(",".data(using: .utf8)!)
```
migs647, line 251: Do we absolutely know 100% of the time this explicit will work on all platforms? Makes me uneasy.

Sources/Segment/Utilities/Storage.swift
```diff
@@ -241,23 +243,20 @@ extension Storage {
...
                 }
-            } catch {
-                assert(false, "Storage: failed to write event to \(storeFile), error: \(error)")
+                // write the data
+                fileHandle?.write(jsonData)
+                if #available(tvOS 13, *) {
```
migs647, line 255: Might be good to add a comment why we are doing this for 13 and above but supporting 11.

Sources/Segment/Utilities/Storage.swift
bsneed: Why would it not work on all platforms??  It's almost identical to what it was before and tests are passing hunky dory.

migs647: I guess it just being a comma should be okay.

migs647 (approved):

PR #62, opened 2021-09-16
Migs647/exampledestination2

migs647:
Adds an ExampleDestination and updates Mixpanel Destination


Segment.xcodeproj/project.pbxproj
```diff
@@ -348,16 +352,17 @@
 		OBJ_11 /* Segment-Tests */ = {
 			isa = PBXGroup;
 			children = (
```
migs647, line 354: I sorted these by kind so they were alphabetical. Because i'm CDO. Which is like OCD but with the letters arranged alphabetically how they should be.

bsneed (approved):

PR #63, opened 2021-09-17
Example of auto notification tracking

bsneed:
- Added example of automatic notification tracking for iOS/tvOS/watchOS.
- Fixed incorrect method signatures in iOS lifecycle plugin.

### DIFF
--- a/Examples/apps/MacExample/MacExample/AppDelegate.swift
+++ b/Examples/apps/MacExample/MacExample/AppDelegate.swift
@@ -22,7 +22,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
             .flushInterval(10)
         
         analytics = Analytics(configuration: configuration)
-
+        //analytics?.add(plugin: NotificationTracking())
     }
 
     func applicationWillTerminate(_ aNotification: Notification) {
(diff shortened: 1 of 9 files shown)


Examples/apps/MacExample/MacExample/AppDelegate.swift
```diff
@@ -22,7 +22,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
             .flushInterval(10)
         
         analytics = Analytics(configuration: configuration)
-
```
migs647, line 25: Should we maybe add a comment: `// Uncomment to add notification tracking to your app` or something?

Sources/Segment/Plugins/Platforms/iOS/iOSLifecycleMonitor.swift
```diff
@@ -10,6 +10,8 @@
 import Foundation
 import UIKit
 
+// NOTE: These method signatures are marked optional as application extensions may not have
```
migs647, line 13: 👍🏻

migs647 (approved):
LGTM

Examples/apps/MacExample/MacExample/AppDelegate.swift
bsneed: It's a placeholder until I get full lifecycle stuff on Mac working.

PR #65, opened 2021-09-21
Fixes issue w/ amplitude sessionId not working

bsneed:
When we changed the application param to be optional for app extensions, this one was missed, so it's falling back to the default implementation that does nothing.


migs647 (approved):
LGTM

dannyhertz:
LGTM :)

PR #76, opened 2021-10-17
update ISO 8601 utility to remove timezone colon

### LINKED ISSUE #75: Incorrect timestamp formats
Looking at the raw events delivered to Segment, I've noticed a discrepancy in the timestamps:

```
...
  -d originalTimestamp=2021-10-16T16:00:27.029:Z \
  -d receivedAt=2021-10-16T16:00:28.078Z \
  -d sentAt=2021-10-16T16:00:27.957:Z \
...
```

You can see that the `originalTimestamp` and `sentAt` dates use a different format than the `receivedAt` date – there's an additional colon prior to the timezone signifier.

When using it with the Amplitude Actions plugin and trying to map the `originalTimestamp` to Amplitude's `timestamp` property this causes the event to be rejected.

By default, the Amplitude Actions destination maps the `receivedAt` property to its timestamp property which of course means events are appearing out of order. It would be much better if we could use the `originalTimestamp` property as currently the events are arriving as shuffled.

tcldr:
The presence of a colon in the ISO 8601 timestamps seems to cause an issue whereby timestamps that rely on the originalTimestamp property, default to the `sentAt` date. I imagine this is due to Segment servers rejecting the malformed timestamp and defaulting to a known value. Removing the colon appears to fix that.

This closes #75


wenxi-zeng:
hi @tcldr can your fetch and merge the upstream commits? this pr should be able to build after that. thanks!

bsneed (approved):

bsneed:
Thanks so much @tcldr !

PR #78, opened 2021-10-19
remove ssh secret

wenxi-zeng:
* remove ssh secret from yml
* remove hacky trick of importing ssh keys from yml
* update dependencies repo from ssh to https

### DIFF
--- a/Segment.xcodeproj/project.pbxproj
+++ b/Segment.xcodeproj/project.pbxproj
@@ -922,7 +922,7 @@
 /* Begin XCRemoteSwiftPackageReference section */
 		462107F02603DA4D00EBC4A8 /* XCRemoteSwiftPackageReference "Sovran-Swift" */ = {
 			isa = XCRemoteSwiftPackageReference;
-			repositoryURL = "git@github.com:segmentio/Sovran-Swift.git";
+			repositoryURL = "https://github.com/segmentio/Sovran-Swift.git";
 			requirement = {
 				kind = upToNextMajorVersion;
 				minimumVersion = 1.0.2;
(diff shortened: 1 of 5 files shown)


Segment.xcodeproj/project.pbxproj
```diff
@@ -922,7 +922,7 @@
 /* Begin XCRemoteSwiftPackageReference section */
 		462107F02603DA4D00EBC4A8 /* XCRemoteSwiftPackageReference "Sovran-Swift" */ = {
 			isa = XCRemoteSwiftPackageReference;
-			repositoryURL = "git@github.com:segmentio/Sovran-Swift.git";
```
migs647, line 925: Is this one absolutely necessary to change for this to work? Apple recommends using ssh vs https for SPM as it can fail in some enterprise environments (this was a big problem at previous employer).

migs647 (commented):
Need answer on changing project.pbxproj. That could be a slippery slope as 1) it can hang up people that don't have https personal access tokens (https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) and 2) this has already been an issue with support tickets. We had some https SPM urls and it was breaking for a user.

Segment.xcodeproj/project.pbxproj
wenxi-zeng: yeah. unfortunately this one is absolutely needed. otherwise, spm uses ssh in the vm, and because of that we have to use a secret to checkout sovran even though it's a public repo.
or we will have to stick `pull_request_target` with the label trick to trigger builds for pr from fork. 😞

migs647: This seems like a really big short coming of the CI build process imo (forcing everything to https). I think we need to have a conversation around this as this is not the path most users will want to take.

wenxi-zeng: agree. it's bad that the https breaks user's build. worst case we can switch back to circle. glad I haven't delete the circle setups.

PR #83, opened 2021-10-25
Made working with and modifying payloads within plugins much simpler.

bsneed:
- Logging is now available outside of an analytics instance.
- Added convenience methods for modifying the integrations/context.
- Added subscripts and methods for JSON value accessing/setting.
- Optimized search order for keypath handlers.
- Unit Tests


Examples/other_plugins/NewRelicErrorReporting.swift
```diff
@@ -0,0 +1,91 @@
...
+    internal let integrationKey = "New Relic"
+    
+    func execute<T: RawEvent>(event: T?) -> T? {
+        // is it a track event?  if not, ignore it.
+        guard let trackEvent = event as? TrackEvent else { return event }
+        // is it the track event we're looking for?  If not, ignore it.
+        guard trackEvent.event == Self.trackName else { return event }
+        // we need to make a new integrations object to keep it from going anywhere
+        // other than new relic
+        guard let newIntegrations = try? JSON([integrationKey: true, "all": false])
```
prayansh: since we have the convenience methods lets use them to showcase the usage :)

Examples/other_plugins/NewRelicErrorReporting.swift
```diff
@@ -0,0 +1,91 @@
...
+// SOFTWARE.
+
+import Foundation
+import Segment
+
+protocol NewRelicConversion {
+    var dictionaryValue: [String: Any]? { get }
+}
+
+class NewRelicErrorReporting: Plugin {
```
prayansh: does newRelic not have any settings?

Sources/Segment/Settings.swift
```diff
@@ -86,6 +86,10 @@ extension Settings: Equatable {
 }
 
 extension Analytics {
+    public func manuallyEnableDestination(plugin: DestinationPlugin) {
```
prayansh, line 93: lets add some docs here about when to use this particular function

Sources/Segment/Utilities/JSON.swift
```diff
@@ -267,19 +268,184 @@ extension JSON {
     }
 }
 
-// MARK: - Mapping
+// MARK: - Mutation
 
 extension JSON {
```
prayansh, line 273: just a note: the kotlin jsonObject is inherently immutable, so the kotlin apis will function a bit differently

Examples/other_plugins/NewRelicErrorReporting.swift
```diff
@@ -0,0 +1,91 @@
...
+        
+        // it's what we're looking for, so lets make sure it *only* goes
+        // to New Relic.
+        trackEvent.integrations = newIntegrations
+        return nil
+    }
+}
+
+extension Analytics {
+    func reportError(_ error: Error, attributes: [String: Any]?) {
```
migs647: Add some documentation here and the next method since these are the public APIs they would call directly.

Sources/Segment/Plugins/Logger/SegmentLog.swift
```diff
@@ -146,16 +163,23 @@ internal extension Analytics {
...
+                    
+                    let log = LogFactory.buildLog(destination: .log, title: "", message: message, kind: filterKind, function: function, line: line)
+                    loggerPlugin.log(log, destination: .log)
                 }
-                
-                let log = LogFactory.buildLog(destination: .log, title: "", message: message, kind: filterKind, function: function, line: line)
-                loggerPlugin.log(log, destination: .log)
             }
+        } else {
+            #if DEBUG
```
migs647, line 179: I'm debating if we should change this from DEBUG to DEBUGLOG, that way our console doesn't get spammed and we can easily disable this and use Console as a backup. Thoughts?

Examples/other_plugins/NewRelicErrorReporting.swift
bsneed: this file wasn't supposed to be here yet.  removed.

Sources/Segment/Utilities/JSON.swift
bsneed: 👍

Examples/other_plugins/NewRelicErrorReporting.swift
bsneed: This file wasn't supposed to be here yet.  Removed.

Sources/Segment/Utilities/JSON.swift
```diff
@@ -267,19 +268,184 @@ extension JSON {
...
         let mapped = try dict.mapTransform(keys, valueTransform: valueTransform)
         let result = try JSON(mapped)
         return result
     }
+    
+    /// Adds a new value to an array and returns a new JSON object.  Function will throw if value cannot be serialized.
+    /// - Parameters:
+    ///   - value: Value to add to the JSON array.
+    ///
+    /// - Returns: A new JSON array with the supplied value added.
```
migs647, line 292: May be worth adding `Throws:` in here and the above method.

Sources/Segment/Utilities/Utils.swift
```diff
@@ -43,3 +43,17 @@ internal func exceptionFailure(_ message: String) {
...
 }
+
+internal protocol Flattenable {
+    func flattened() -> Any?
+}
+
+extension Optional: Flattenable {
+    internal func flattened() -> Any? {
+        switch self {
+        case .some(let x as Flattenable): return x.flattened()
```
migs647, line 54: v00d00

Sources/Segment/Settings.swift
bsneed: 👍

Tests/Segment-Tests/JSON_Tests.swift
```diff
@@ -72,19 +84,52 @@ class JSONTests: XCTestCase {
...
+        XCTAssertNotNil(jsonObject2)
+        
+        let structValue: TestStruct? = jsonObject[keyPath: "test"]
+        XCTAssertEqual(structValue!.str, "hello")
+
+        let intValue: Int? = jsonObject[keyPath: "test.dict.1"]
+        XCTAssertEqual(intValue, 1)
+        
+        jsonObject[keyPath: "structSet.object"] = test
+        jsonObject[keyPath: "codys.brain.melted"] = true
```
migs647, line 120: 🤯

Tests/Segment-Tests/JSON_Tests.swift
```diff
@@ -72,19 +84,52 @@ class JSONTests: XCTestCase {
...
+        
+        XCTAssertTrue(jsonObject[keyPath: "structSet.object.str"] == "hello")
+        XCTAssertTrue(jsonObject[keyPath: "codys.brain.melted"] == true)
```
migs647, line 123: 🤯👍🏻

Tests/Segment-Tests/KeyPath_Tests.swift
```diff
@@ -189,14 +202,42 @@ class KeyPath_Tests: XCTestCase {
         XCTAssertTrue(dict["user_id"] as? String == "brandon")
     }
     
+    // useful for once-in-awhile checking, but doesn't need to be run as part of
+    // the overall test suite.
+    /*func testDictDeconstructionSpeed() {
```
migs647, line 207: Do we want to leave this commented code in?

migs647 (approved):
LGTM. Just maybe some documentation and remove some commented tests.

Tests/Segment-Tests/KeyPath_Tests.swift
bsneed: Yes; While we don't need to do perf measurements all the time for this IMO.  I don't expect to change much/ever.  But it is nice to have should we want to test it at some point.

Sources/Segment/Plugins/Logger/SegmentLog.swift
bsneed: I'm not really following what you mean here.  Please hit me up on slack.

PR #88, opened 2021-10-28
Fix JSON Error on archive

schenkty:
Currently, the latest version fails to be archived due to the throw type being a `String`. This PR resolves the issue by leveraging the `JSONError` type.


bsneed:
Thanks @schenkty!  We addressed this in a slightly different way in #90 along with removing the String:Error conformance that was leaking in from Sovran-Swift.  Was released just now as 1.0.5

PR #92, opened 2021-11-05
Integration enable/disable refactor

bsneed:
- Modified timeline to determine enablement by a combination of event data as well as settings data.
- Added method `hasIntegrationSettings`.
- Removed confusing state entries re integrations vs. settings for integrations.
- Modified Segment destination to construct integration object as it's being sent to segment rather than try to do it beforehand.

### DIFF
--- a/Sources/Segment/Analytics.swift
+++ b/Sources/Segment/Analytics.swift
@@ -24,6 +24,9 @@ public class Analytics {
     
     public var timeline: Timeline
     
+    /// Initialize this instance of Analytics with a given configuration setup.
+    /// - Parameters:
+    ///    - configuration: The configuration to use
     public init(configuration: Configuration) {
         self.configuration = configuration
         
@@ -44,7 +47,7 @@ public class Analytics {
         _ = timeline.process(incomingEvent: event)
     }
     
-    public func process(event: RawEvent) {
+    internal func process(event: RawEvent) {
         switch event {
         case let e as TrackEvent:
             timeline.process(incomingEvent: e)
@@ -65,6 +68,7 @@ public class Analytics {
 // MARK: - System Modifiers
 
 extension Analytics {
+    /// Returns the anonymousId currently in use.
     public var anonymousId: String {
         if let userInfo: UserInfo = store.currentState() {
             return userInfo.anonymousId
@@ -72,6 +76,7 @@ extension Analytics {
         return ""
     }
     
+    /// Returns the userId that was specified in the last identify call.
     public var userId: String? {
         if let userInfo: UserInfo = store.currentState() {
             return userInfo.userId
@@ -79,6 +84,7 @@ extension Analytics {
         return nil
     }
     
+    /// Returns the traits that were specified in the last identify call.
     public func traits<T: Codable>() -> T? {
         if let userInfo: UserInfo = store.currentState() {
             return userInfo.traits?.codableValue()
@@ -86,6 +92,8 @@ extension Analytics {
         return nil
     }
     
+    /// Tells this instance of Analytics to flush any queued events up to Segment.com.  This command will also
+    /// be sent to each plugin present in the system.
     public func flush() {
         apply { plugin in
             if let p = plugin as? EventPlugin {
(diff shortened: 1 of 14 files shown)


Sources/Segment/Settings.swift
```diff
@@ -68,12 +68,9 @@ public struct Settings: Codable {
         return integrationSettings(forKey: plugin.key)
     }
     
-    public func isDestinationEnabled(key: String) -> Bool {
+    public func hasIntegrationSettings(forPlugin plugin: DestinationPlugin) -> Bool {
```
migs647, line 71: I like this rename

Sources/Segment/Analytics.swift
```diff
@@ -111,6 +111,15 @@ extension Analytics {
         return settings
     }
     
+    /// Manually enable a destination plugin.  This is useful when a given DestinationPlugin doesn't have any Segment tie-ins at all.
+    /// This will allow the destination to be processed in the same way within this library.
+    /// - Parameters:
+    ///   - plugin: The destination plugin to enable.
+    public func manuallyEnableDestination(plugin: DestinationPlugin) {
```
migs647, line 147: I'm going back and forth if we want this to be a "core" public API. It makes sense for this to be separated by an extension to me. Thoughts?

Sources/Segment/Timeline.swift
```diff
@@ -210,18 +210,28 @@ extension DestinationPlugin {
...
+        if let disabled: Bool = integrations?.value(forKeyPath: KeyPath(self.key)), disabled == false {
+            customerDisabled = true
+        }
+        
+        var hasSettings = false
+        if let settings = analytics?.settings() {
+            hasSettings = settings.hasIntegrationSettings(forPlugin: self)
+        }
+        
+        return (hasSettings && customerDisabled == false)
```
migs647: I'd prefer this line as `hasSettings == true && customerDisabled == false` OR `hasSettings && !customerDisabled` so we aren't mixing semantics.

migs647:
Liking this direction

oscb (approved):
Looks good to me!

alanjcharles (approved):
my brain hurts but it looks good and make sense

Sources/Segment/Plugins/SegmentDestination.swift
```diff
@@ -156,6 +140,29 @@ public class SegmentDestination: DestinationPlugin {
...
 
+// MARK: - Utility methods
+extension SegmentDestination {
+    internal func configureCloudDestinations<T: RawEvent>(event: T) -> T {
+        guard let integrationSettings = analytics?.settings() else { return event }
+        guard let plugins = analytics?.timeline.plugins[.destination]?.plugins as? [DestinationPlugin] else { return event }
+        guard let customerValues = event.integrations?.dictionaryValue else { return event }
+        
+        // take the customer values first.
+        var merged = customerValues
```
prayansh: i think we should use integrationSettings as the base value, and then add the customer values to override here

Sources/Segment/Timeline.swift
```diff
@@ -210,18 +210,28 @@ extension DestinationPlugin {
         }
         return result
     }
+    
+    internal func isDestinationEnabled(integrations: JSON?) -> Bool {
```
prayansh: maybe we should change the input parameter to `RawEvent` so that the api isnt misused with any kind of JSON

prayansh (approved):
looks good. Added some comments for more clarification

Sources/Segment/Types.swift
```diff
@@ -277,14 +277,13 @@ extension RawEvent {
...
         
-        guard let system: System = store.currentState() else { return self }
         guard let userInfo: UserInfo = store.currentState() else { return self }
         
         result.anonymousId = userInfo.anonymousId
         result.userId = userInfo.userId
         result.messageId = UUID().uuidString
         result.timestamp = Date().iso8601()
-        result.integrations = system.integrations
+        result.integrations = try? JSON([String: Any]())
```
oscb, line 286: Don't you need to leave the `self.integrations` here? Else it strips all user defined integrations before the timeline processes it

bsneed: Yeah; this happens right when the event comes in.  The user hasn't had a chance to do anything yet so there's nothing to override.

PR #93, opened 2021-11-11
fix: screen event needs to be setup properly

pietropizzi:
The screen call would be calling to the wrong process method so that the event would not be set up correctly. In my case the `integrations` were not setup correctly so when I used `enableCloudIntegration` on the rawEvent it would throw an error complaining that integrations is not a dictionary.

I ran all the tests and they are passing. I did not find any tests specifically for this behaviour and was not familiar enough with the codebase to do that myself.

### DIFF
--- a/Sources/Segment/Events.swift
+++ b/Sources/Segment/Events.swift
@@ -188,7 +188,7 @@ extension Analytics {
                 exceptionFailure("Could not parse properties.")
             }
         }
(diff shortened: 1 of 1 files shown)


bsneed (approved):

bsneed:
thanks @pietropizzi !  I found another instance of this in group as well.  Nice catch!

PR #98, opened 2021-11-17
use cache directory instead of documents

rhaining:
i kept seeing an error that tvOS was having trouble writing to anything inside the Documents directory. 

it sounds like for tvOS, they suggest using Cache for anything you need:
https://developer.apple.com/library/archive/documentation/General/Conceptual/AppleTV_PG/index.html#//apple_ref/doc/uid/TP40015241
https://developer.apple.com/forums/thread/89008

### DIFF
--- a/Sources/Segment/Utilities/Storage.swift
+++ b/Sources/Segment/Utilities/Storage.swift
@@ -182,7 +182,7 @@ extension Storage {
     }
     
     private func eventStorageDirectory() -> URL {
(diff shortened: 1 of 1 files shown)


bsneed:
Merged #103 in lieu of this PR.

bsneed:
Thanks @rhaining for bringing this to our attention!  The linked PR should address the issue in a less disruptive way for other platforms as well as give us some room to expand the exclusions if necessary.

PR #103, opened 2022-01-05
Use different path for tvOS storage.

bsneed:
Alternate way of accepting PR #98.  We may have to make further OS distinctions in the future.


migs647 (approved):
YAY

PR #108, opened 2022-01-12
on macOS, like on tvOS use the caches folder

diederich:
On macOS, if the app is not sandboxed, the folder for events storage ends up in the user's `~/Documents/` directory, which potentially gets synced via iCloud Documents.

Let's also use the caches directory (`~/Library/Caches/`) on macOS, same as tvOS.

(I had a previous patch which moved all platforms to caches directory, given the documents folder on iOS might also be synced across devices. WDYT?)


bsneed (approved):

bsneed:
Great catch @diederich !  Thanks so much!  Will go out with the next release.

PR #114, opened 2022-02-09
correct identify doc

wenxi-zeng:
* correct identify docs
* add badges


wenxi-zeng:
removed codecov changes, since codecov v1 uses bash uploader which has security issue and v2 does not support `xcresult` format yet. a workaround for v2 is to use a script to convert `xcresult` to `json` format, which is not very ideal. we can come back to this later.

see issue from codecov:
codecov/uploader#223

bsneed (approved):

PR #115, opened 2022-02-13
Fix cleanup handler not called on background task expiration due to struct being captured before assignment

davisg123:
We were encountering this issue:
https://github.com/segmentio/analytics-swift/issues/54

The PR for the issue (https://github.com/segmentio/analytics-swift/pull/111) does prevent background crashes due to the upload task now completing on time

However, there is still an issue; `application.endBackgroundTask` is not called for tasks that do happen to expire in the background

`UploadTaskInfo` as a value type is captured by `beginBackgroundTask` before the `cleanup` handler is assigned. You can see the issue by running version 1.0.6 where tasks expired often and set the breakpoint as seen in the attached screenshot

My fix is to no longer capture self in either closure.

Screen Shot 2022-02-12 at 4 31 44 PM

### DIFF
--- a/Sources/Segment/Plugins/Platforms/iOS/iOSLifecycleMonitor.swift
+++ b/Sources/Segment/Plugins/Platforms/iOS/iOSLifecycleMonitor.swift
@@ -190,14 +190,15 @@ extension SegmentDestination.UploadTaskInfo {
         self.task = task
         
         if let application = UIApplication.safeShared {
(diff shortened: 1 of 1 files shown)


bsneed:
Thanks @davisg123 !  Just merged.  I'm gonna do another PR to remove taskID since it's no longer needed.  Great catch!

PR #125, opened 2022-03-14
fix: fix timing issue in idfa

alanjcharles:
fixes timing issue that would result in tracking status remaining `.notDetermined` and loop in `askForPermission()`

### DIFF
--- a/Examples/other_plugins/IDFACollection.swift
+++ b/Examples/other_plugins/IDFACollection.swift
@@ -79,8 +79,9 @@ extension IDFACollection: iOSLifecycle {
         let status = ATTrackingManager.trackingAuthorizationStatus
         if status == .notDetermined && !alreadyAsked {
             // we don't know, so should ask the user.
-            askForPermission()
             alreadyAsked = true
+            askForPermission()
+           
         }
     }
 }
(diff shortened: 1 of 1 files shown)


Examples/other_plugins/IDFACollection.swift
```diff
@@ -77,6 +74,17 @@ class IDFACollection: Plugin {
     }
 }
 
+extension IDFACollection: iOSLifecycle {
+    func applicationDidBecomeActive(application: UIApplication?) {
+        let status = ATTrackingManager.trackingAuthorizationStatus
+        if status == .notDetermined && !alreadyAsked {
+            // we don't know, so should ask the user.
+            askForPermission()
+            alreadyAsked = true
```
bsneed, line 82: lets move alreadyAsked=true above askForPermission().

bsneed (changes requested):

Examples/other_plugins/IDFACollection.swift
alanjcharles: @bsneed so remove the new extension, add `iOSLifecycle` protocol to existing `extension IDFA` as well?

bsneed (approved):

PR #130, opened 2022-03-28
Revert App Opened change from #101

bsneed:


migs647 (approved):
LGTM -- did just realize we don't have tests in analytics-swift for foregrounding.

PR #132, opened 2022-04-06
Added ability to output XCFramework via Build.sh.

migs647:
Needed to change build setting in Xcode. Build Library For Distribution needed to be Yes


migs647:
This won't be ready until Xcode 13.3. Closing.

PR #142, opened 2022-05-06
Added convenience methods for ObjC interop

bsneed:
- Added convenience methods for ObjC interop.
- Reduced cellular info example down to just iOS to fix CI issue.

### DIFF
--- a/Examples/apps/SegmentUIKitExample/SegmentUIKitExample/AppDelegate.swift
+++ b/Examples/apps/SegmentUIKitExample/SegmentUIKitExample/AppDelegate.swift
@@ -18,7 +18,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         Analytics.main.add(plugin: IDFACollection())
         Analytics.main.add(plugin: UIKitScreenTracking())
         Analytics.main.add(plugin: NotificationTracking())
+        
+#if os(iOS) && !targetEnvironment(macCatalyst)
         Analytics.main.add(plugin: CellularCarrier())
+#endif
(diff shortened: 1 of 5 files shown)


Examples/apps/SegmentUIKitExample/SegmentUIKitExample/AppDelegate.swift
```diff
@@ -18,7 +18,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         Analytics.main.add(plugin: IDFACollection())
         Analytics.main.add(plugin: UIKitScreenTracking())
         Analytics.main.add(plugin: NotificationTracking())
+        
+#if os(iOS) && !targetEnvironment(macCatalyst)
         Analytics.main.add(plugin: CellularCarrier())
```
migs647, line 23: 👍

Sources/Segment/ObjC/ObjCAnalytics.swift
```diff
@@ -13,12 +13,20 @@ import Foundation
 
 @objc(SEGAnalytics)
 public class ObjCAnalytics: NSObject {
-    internal let analytics: Analytics
+    /// The underlying Analytics object we're working with
+    public let analytics: Analytics
```
migs647, line 17: We had to expose this for tests?

bsneed: makes cross-config way easier if you can init in objc and then access it later in swift if needed and vice versa.

PR #148, opened 2022-07-11
Additions to enable CLI support

bsneed:
- Added property to determine if there are unsent events.
- Made some properties public for CLI visibility.
- Improved file handle access and queue determinations.
- Improved decode failure error message.
- Fixed cellular plugin key path.


prayansh (approved):

Sources/Segment/Plugins/SegmentDestination.swift
```diff
@@ -124,6 +124,7 @@ public class SegmentDestination: DestinationPlugin {
                     switch result {
                         case .success(_):
                             storage.remove(file: url)
+                            self.cleanupUploads()
```
grzegorzkrukowski, line 127: Hey @bsneed  - why this line has been added here when there is another call of this method that happens independently from the result value?

https://github.com/segmentio/analytics-swift/pull/148/files#diff-92679f628b3def3ba245e4435edfcea09ea06a77277e6e2497aa331efbf46028R136

bsneed: Nice catch @grzegorzkrukowski !  The cleanupUploads call here isn't necessary.  Fortunately the subsequent call results in a no-op.

Issue #152, opened 2022-08-11
SegmentDestination is causing huge CPU spikes when blocked by LittleSnitch


grzegorzkrukowski:
**Describe the bug**
When using LittleSnitch to block traffic to Segement.io from our application, we are getting huge CPU Spikes really often, and when entering foreground/background -  basically every time the `flush` is called.

I think there are several problems in `SegmentDestination`:
1. Storage can grow without limits - old events that are not sent are never removed.
2. Storage is keeping files infinitely - which is causing a snowball effect for every sync.
3. `Flush` logic is not well prepared for having a lot of data to work with.
First of all, it read all the data from the filesystem every time it's called.
Then for each of the files, it creates a network request straight away iterating over all of them - which in my case ends up being over 1500 requests created and failing in milliseconds. This is when CPU spikes are kicking in.
The whole process is repeated every time the `flush()` is called and things are getting worse and worse over time when more un-send events are piling up in the storage.

**To Reproduce**
1. Install `LittleSnitch` 
2. Block segment.io traffic in `LittleSnitch`
3. Create around 100.000 tracked events in Segment
4. Observe the CPUSpikes when `flush` logic is called.
5. Log an enormous amount of network connections created during the `flush()` call.

**Expected behavior**
1. If the user decides to block segment.io coming from our application using external software there should be no performance drop or CPU spikes coming out of this library.

2. Storage should not grow infinitely - at some point, older events should be removed when newer events are added - there should be the maximum amount of files that are allowed to be cached for sending.

3. `flush()` method should not go so aggressive with creating network events for all un-send files - instead it should either go one by one and read the next files only when the previous one succeeded or it should pre-check if segment batch API is reachable before trying to create a massive amount of requests.

**Platform (please complete the following information):**
 - Library Version in use: 1.2.2
 - Platform being tested: macOS
 - Integrations in use: Amplitude

**Workaround**
As for now, I am using an ugly workaround from [this fork](https://github.com/grzegorzkrukowski/analytics-swift/blob/14c9c9f6e8cff4d173ba214c53479c0dc411d02c/Sources/Segment/Plugins/SegmentDestination.swift#L129-L131) that simply deleted all failed events if the error is `URLError.networkConnectionLost` - which is the case for LittleSnitch.

bsneed:
Looking at this today.

bsneed:
@grzegorzkrukowski could you have a look at or maybe try the PR above and see if it resolves the issue for you?  Thanks!

[...]

PR #153, opened 2022-08-11
Fix issue where cascading network timeouts compound on each other

bsneed:
Fixes issue #152 

It was observed that when network traffic is blocked that outbound requests are timing out after 60 seconds, however the flush interval has a default of 30 seconds.  This would result in a cascading set of outbound network requests that would all fail.

This change makes it such that in the case where connectivity is unavailable, we pause the flush timer for 5 minutes before attempting again.

### DIFF
--- a/Sources/Segment/Plugins/SegmentDestination.swift
+++ b/Sources/Segment/Plugins/SegmentDestination.swift
@@ -120,20 +120,27 @@ public class SegmentDestination: DestinationPlugin {
             for url in data {
                 analytics.log(message: "Processing Batch:\n\(url.lastPathComponent)")
                 
-                let uploadTask = httpClient.startBatchUpload(writeKey: analytics.configuration.values.writeKey, batch: url) { (result) in
+                let uploadTask = httpClient.startBatchUpload(writeKey: analytics.configuration.values.writeKey, batch: url) { [weak self] (result) in
                     switch result {
                         case .success(_):
                             storage.remove(file: url)
-                            self.cleanupUploads()
-                        default:
-                            analytics.logFlush()
+                            analytics.log(message: "Processed: \(url.lastPathComponent)")
+                        case .failure(let e):
+                            analytics.log(message: "Failed to process: \(url.lastPathComponent), \(e)")
+                            if e.connectivityError {
+                                self?.flushTimer?.suspend()
+                                DispatchQueue.main.asyncAfter(deadline: .now() + 300) { [weak self] in
+                                    self?.flushTimer?.resume()
+                                }
+                            }
                     }
                     
-                    analytics.log(message: "Processed: \(url.lastPathComponent)")
+                    analytics.logFlush()
+
                     // the upload we have here has just finished.
                     // make sure it gets removed and it's cleanup() called rather
                     // than waiting on the next flush to come around.
-                    self.cleanupUploads()
+                    self?.cleanupUploads()
                 }
                 // we have a legit upload in progress now, so add it to our list.
                 if let upload = uploadTask {
--- a/Sources/Segment/Plugins/SegmentDestination.swift
+++ b/Sources/Segment/Plugins/SegmentDestination.swift
@@ -128,6 +128,8 @@ public class SegmentDestination: DestinationPlugin {
                         case .failure(let e):
                             analytics.log(message: "Failed to process: \(url.lastPathComponent), \(e)")
                             if e.connectivityError {
+                                // In the event of a connectivity error, suspend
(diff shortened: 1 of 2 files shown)


prayansh (approved):
LGTM 🚢

brianmichel:
Hey folks, any ETA on when this will land? We're going to be adding a work around to our fork as well, but would love to pick up the main line solution when possible, thanks!

eco-travis:
@prayansh Any idea on when this might get merged?  This PR seems to be getting pretty stale.

bclous:
@bsneed, any reason why this hasn't been merged? This is causing major issues in our app. It seems anyone with an ad blocker is getting a CPU spike within 45 seconds and crashing.

brianmichel:
> @bsneed, any reason why this hasn't been merged? This is causing major issues in our app. It seems anyone with an ad blocker is getting a CPU spike within 45 seconds and crashing.

Hey @bclous dunno if you need a quick solution, we ended up adding an API that allows you to only keep a maximum number of files on disk to ensure that we wouldn't accumulate too many files and cause all kinds of issues for certain folks. It's here if you want it https://github.com/thebrowsercompany/analytics-swift/pull/1 should be straightforward to cherry-pick into your own fork.

bclous:
> > @bsneed, any reason why this hasn't been merged? This is causing major issues in our app. It seems anyone with an ad blocker is getting a CPU spike within 45 seconds and crashing.
> 
> Hey @bclous dunno if you need a quick solution, we ended up adding an API that allows you to only keep a maximum number of files on disk to ensure that we wouldn't accumulate too many files and cause all kinds of issues for certain folks. It's here if you want it [thebrowsercompany#1](https://github.com/thebrowsercompany/analytics-swift/pull/1) should be straightforward to cherry-pick into your own fork.

Thanks for the reply. Is this only in the browser company's fork? and Not in the current 1.3.2 version?

bsneed:
@bclous @brianmichel the original reporter of the bug suggested it didn't go far enough.  The attached bug report has more of the conversation.  We have a solution, but haven't had the bandwidth to implement just yet.  The reporter of the issue had a fork with a temporary workaround that got him going again.  It may be worth it to check that out in the meantime.  If you have a Segment account, reaching out directly to support will help get this prioritized.

bclous:
> @bclous @brianmichel the original reporter of the bug suggested it didn't go far enough. The attached bug report has more of the conversation. We have a solution, but haven't had the bandwidth to implement just yet. The reporter of the issue had a fork with a temporary workaround that got him going again. It may be worth it to check that out in the meantime. If you have a Segment account, reaching out directly to support will help get this prioritized.

Got it, thanks for all the info

bsneed:
@bclous as a follow up; I just made the case in our sprint planning to get this done.  I've assigned an engineer to take a first stab at it on another platform and will bring that over once he's got the kinks worked out.  The idea is we'd implement a flush policy that developers could customize to gain more control over how flushing works.

The idea is that for your particular use case(s) you would be able to implement a policy that could do this like ...

example 1: network requests are failing, so back off on the flush interval.

example 2: event storage is piling up, so start culling files.

etc.

bclous:
@bsneed thanks, I think that makes sense. We had an extremely similar problem in our custom infra at FB. Unsent events were piling up on disk and causing all sorts of performance issues. 

Would be great to have greater control over flushing and which events should absolutely be stored versus events that we're fine dropping if need be.

brianmichel:
> > > @bsneed, any reason why this hasn't been merged? This is causing major issues in our app. It seems anyone with an ad blocker is getting a CPU spike within 45 seconds and crashing.
> > 
> > 
> > Hey @bclous dunno if you need a quick solution, we ended up adding an API that allows you to only keep a maximum number of files on disk to ensure that we wouldn't accumulate too many files and cause all kinds of issues for certain folks. It's here if you want it [thebrowsercompany#1](https://github.com/thebrowsercompany/analytics-swift/pull/1) should be straightforward to cherry-pick into your own fork.
> 
> Thanks for the reply. Is this only in the browser company's fork? and Not in the current 1.3.2 version?

Yup, this is our fork which is up to date with the 1.3.1 release from Segment. I can merge upstream if you needed something specific from 1.3.2, however we also specifically omit tracking information so I don't think you should take it wholesale without taking a peek through the changes that are unique to our fork, hence suggesting you fork and cherry-pick where needed.

bsneed:
Fixed in 1.4.4 via Flush Policy and an example flush policy implementation here: https://github.com/segmentio/analytics-swift/blob/main/Examples/tasks/NetBlockerFlushPolicy.swift

PR #161, opened 2022-10-06
Fix nil value parsing

VolodymyrHanas:
Comparing `Any` to `nil` always returns false, so when nil is passed as a value in traits you will always receive a `nonJSONType` error


bsneed:
Looks like the test runner is misconfigured for PRs; will test locally.

bsneed (approved):

bsneed:
Nice find @VolodymyrHanas !  Thanks so much for the PR! ✋

Issue #162, opened 2022-10-14
[Segment ERROR - finish(file:):320] Unable to rename to temp warning/error


malhotrachetan:
**Describe the bug**

Coming across a very weird debug message lately since the last update. The message appears around 20 secs post starting the app. The message says:

`[Segment ERROR - finish(file:):320]
Unable to rename to temp: file:///Users/chetanmalhotra/Library/Developer/CoreSimulator/Devices/094C9BEE-0AB5-4345-AB1D-2CA907B66033/data/Containers/Data/Application/1688B1EE-1CCC-44B7-AA72-FB7A9A272B3E/Documents/segment/XVFH4DGT0ufXKfxlMQzNMQDosdD9eKLW/7-segment-events, Error: Error Domain=NSCocoaErrorDomain Code=4 "“7-segment-events” couldn’t be moved to “XVFH4DGT0ufXKfxlMQzNMQDosdD9eKLW” because either the former doesn’t exist, or the folder containing the latter doesn’t exist." UserInfo={NSSourceFilePathErrorKey=/Users/chetanmalhotra/Library/Developer/CoreSimulator/Devices/094C9BEE-0AB5-4345-AB1D-2CA907B66033/data/Containers/Data/Application/1688B1EE-1CCC-44B7-AA72-FB7A9A272B3E/Documents/segment/XVFH4DGT0ufXKfxlMQzNMQDosdD9eKLW/7-segment-events, NSUserStringVariant=(
    Move
), NSDestinationFilePath=/Users/chetanmalhotra/Library/Developer/CoreSimulator/Devices/094C9BEE-0AB5-4345-AB1D-2CA907B66033/data/Containers/Data/Application/1688B1EE-1CCC-44B7-AA72-FB7A9A272B3E/Documents/segment/XVFH4DGT0ufXKfxlMQzNMQDosdD9eKLW/7-segment-events.temp, NSFilePath=/Users/chetanmalhotra/Library/Developer/CoreSimulator/Devices/094C9BEE-0AB5-4345-AB1D-2CA907B66033/data/Containers/Data/Application/1688B1EE-1CCC-44B7-AA72-FB7A9A272B3E/Documents/segment/XVFH4DGT0ufXKfxlMQzNMQDosdD9eKLW/7-segment-events, NSUnderlyingError=0x600001e34e40 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}`

**To Reproduce**
[...]

Issue #154, opened 2022-09-08
Application is crashing on Storage errors


grzegorzkrukowski:
**Describe the bug**

We are getting crash reports coming from the Segment library. 
According to crash logs, it's related to problems with writing files by a Storage class.

The problem seems to be related to using some deprecated methods of FileHandler class, which are throwing uncatchable errors:
```
extension FileHandle {    
    /* The API below may throw exceptions and will be deprecated in a future version of the OS.
     Use their replacements instead. */
    
    @available(macOS, introduced: 10.0, deprecated: 100000)
    open func write(_ data: Data)
```

**To Reproduce**
1. Try to send any Segment event when disc has no more space left

**Expected behavior**
The segment should not cause application crash.
Ideally, the library should allow us to read those errors so we can react to them.

**Platform (please complete the following information):**
 - Library Version in use: 1.3.1
 - Platform being tested: macOS
 - Integrations in use: Amplitude

**Additional context**
Here is the important part of the stack trace from a crash:
```
OS Version: macOS 12.5.1 (21G83)
Report Version: 104

Exception Type: EXC_CRASH (SIGABRT)
Crashed Thread: 0

Application Specific Information:
*** -[NSConcreteFileHandle writeData:]: No space left on device

Thread 0 Crashed:
0   CoreFoundation                  0x315d69194         __exceptionPreprocess
1   libobjc.A.dylib                 0x3158eae00         objc_exception_throw
2   Foundation                      0x317bef54c         _NSFileHandleRaiseOperationExceptionWhileReading
3   Foundation                      0x317ac0c40         -[NSConcreteFileHandle writeData:]
4   Raycast                         0x205295b54         Storage.storeEvent (Storage.swift:268)
5   Raycast                         0x20529509c         Storage.write<T> (Storage.swift:37)
6   Raycast                         0x205265178         thunk for closure
7   Raycast                         0x205265198         thunk for closure
8   libdispatch.dylib               0x3158481b0         _dispatch_client_callout
9   libdispatch.dylib               0x315857410         _dispatch_lane_barrier_sync_invoke_and_complete
10  Raycast                         0x20529433c         Storage.write<T> (Storage.swift:32)
11  Raycast                         0x205265984         SegmentDestination.queueEvent<T> (SegmentDestination.swift:99)
12  Raycast                         0x205265858         SegmentDestination.execute<T> (SegmentDestination.swift:78)
[...]

PR #163, opened 2022-10-15
Storage hardening; Add error handling capabilities.

bsneed:
- Added OutputFileStream class to shield us a little from Swift's back-and-forth/deprecations on FileHandle.
- Added storage stress tests to test suite.
- Improved some existing test cases.
- Added `errorHandler` to configuration so developer can receive errors we see internally.

Affects #154 and #162.


alanjcharles (approved):

wenxi-zeng (approved):

Sources/Segment/Configuration.swift
```diff
@@ -95,5 +99,16 @@ public extension Configuration {
         values.cdnHost = value
         return self
     }
+    
+    @discardableResult
+    func errorHandler(_ value: @escaping (Error) -> Void) -> Configuration {
```
oscb, line 101: Just curious as to why it returns the `Configuration` ?

Sources/Segment/Errors.swift
```diff
@@ -0,0 +1,79 @@
...
+                return AnalyticsError.jsonUnableToDeserialize(e)
+            case .unknown:
+                return AnalyticsError.jsonUnknown(e)
+            }
+        }
+        return error
+    }
+    
+    /// Reports an internal error to the user-defined error handler.
+    public func reportInternalError(_ error: Error, hardStop: Bool = false) {
```
oscb: nit: maybe `critical` instead of `hardStop` sounds more exception-like?

Sources/Segment/Plugins/SegmentDestination.swift
```diff
@@ -48,15 +49,23 @@ public class SegmentDestination: DestinationPlugin {
...
         guard let analytics = self.analytics else { return }
         storage = analytics.storage
         httpClient = HTTPClient(analytics: analytics)
-        flushTimer = QueueTimer(interval: analytics.configuration.values.flushInterval) { [weak self] in
-            self?.flush()
+        
+        // flushInterval and flushAt can be modified post initialization
+        analytics.store.subscribe(self, initialState: true) { [weak self] (state: System) in
+            guard let self = self else { return }
+            self.flushTimer = QueueTimer(interval: state.configuration.values.flushInterval) { [weak self] in
```
oscb, line 63: Not sure how this works in Swift but wouldn't it also have to cancel the previous timer?

oscb (approved):
Left a few questions and small comments, but overall I think this looks good!

Sources/Segment/Plugins/SegmentDestination.swift
bsneed: QueueTimer is a wrapper on a dispatch source timer.  Cancel is handled in dealloc.

Sources/Segment/Errors.swift
bsneed: Good call.

Sources/Segment/Configuration.swift
bsneed: The config methods are all chainable and this continues that.  Makes is so you can do ...

```
let analytics = Configuration(writeKey: "1234")
                         .trackLifecycleEvents(true)
                         .apiHost("eu1.segmentapi.com")
                         .whateverElse()
```

.. instead of having to create an instance and then set them all afterwards.

Sources/Segment/Errors.swift
bsneed: named it `fatal` instead as it seemed more descriptive of what was going to happen, ie: YOUR APP WILL DIE.

PR #164, opened 2022-10-17
Fix unnecessary retains to prevent zombie Analytics instances

### LINKED ISSUE #159: Zombie Analytics after deallocation.
**Describe the bug**
Creating and deallocating the Analytics instances doesn't stop them from sending events to the back-end sources. 

**To Reproduce**
1. Allocate Analytics instance (1) with support for lifecycle events and connected to the Segment source.
2. Background the app and watch for application lifecycle event to appear in Segment debugger.
3. Deallocate analytics instance (1) and allocate Analytics instance (2) with support for lifecycle events and connected to the same Segment source.
4. Background the app and watch for duplicate application lifecycle events to appear in Segment debugger.
5. Goto 3. for more fun and more events!

**Expected behavior**
The Analytics instance should clean up after itself and shutdown its plugins when deallocated.   To be clear, I'm not expecting two valid instances to coordinate in any way - that would reasonably produce duplicates, but if there is only one app-wide, live instance that hasn't been released, it should be the _only_ one sending events.

**Screenshots**
image

 - iOS library version 1.3.0

**Additional context**
My use case requires my app be able to reconfigure Segment between logins so that I can attach to QA and production depending on the user's preference.

bsneed:
- Fixed a few retain loops.
- Added tests to capture leaks in the future.
- Did some cleanup as necessary.

Fixes #159

### DIFF
--- a/Examples/destination_plugins/AdjustDestination.swift
+++ b/Examples/destination_plugins/AdjustDestination.swift
@@ -41,7 +41,7 @@ class AdjustDestination: NSObject, DestinationPlugin, RemoteNotifications {
     let timeline = Timeline()
     let type = PluginType.destination
     let key = "Adjust"
-    var analytics: Analytics? = nil
+    weak var analytics: Analytics? = nil
     
     private var settings: AdjustSettings? = nil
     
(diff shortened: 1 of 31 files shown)


bsneed:
I see I missed some NotificationCenter ones; fixing...

MichaelGHSeg (approved):

Issue #174, opened 2022-11-08
The `os.name` context is reported incorrectly on macOS


seanrucker:
The `MacOSVendorSystem` uses `device.operatingSystemVersionString` for the `systemName` property. I believe this is incorrect as it is resulting in values such as "Version 12.5.1 (Build 21G83)" being reported as the name of the operating system. On iOS it uses `device.systemName` which results in "iOS" being reported as the name of the operating system, which is what I would expect.

**To Reproduce**
1. Implement the SDK in a macOS app.
2. Run the app then background it to flush the events.
3. Check the Segment debugger and notice the `os` context values look like the following.

```json
"os": {
    "name": "Version 12.5.1 (Build 21G83)",
    "version": "12.5.1"
}
```

**Expected behavior**
I expect the value for `os.name` on macOS to be "macOS".

**Screenshots**
![CleanShot 2022-11-08 at 10 56 35@2x](https://user-images.githubusercontent.com/188337/200613601-3f853e85-185c-448e-afb5-22680490ed44.png)

**Platform (please complete the following information):**
 - Library Version in use: 1.3.2
 - Platform being tested: macOS
 - Integrations in use: None

alanjcharles:
Hi @seanrucker! It seems Apple deprecated `operatingSystem()` and `operatingSystemName()` awhile back for macOS and this is the correct method to return the system name. However, I take your point about the stuff in parens, let me get back to you on that. 
<img width="571" alt="image" src="https://user-images.githubusercontent.com/50601149/200686851-84bb8f62-cad0-4b2f-aed7-cc272f4714e6.png">

seanrucker:
Maybe it makes sense just to hardcode it to "macOS"? Even without the parenthesis in the name, it's hard to work with. The name "Version 12.5.1" doesn't tell me that it's coming from a macOS device and makes it difficult to filter in analytics tools.

On iOS it simply reports "iOS" for the name and on iPadOS it reports "iPadOS" which is what I would expect. 

I want to be able to produce reports in our analytics tools that show the differences in the three platforms and if `os.name` isn't consistent between OS versions, then we can't use it that way. Seeing as there is a separate `os.version` property, it seems to me the name value should not change.

PR #176, opened 2022-11-09
hardcode `macOS` value in context.os.name

alanjcharles:
- hardcodes `macOS` in `context.os.name`
- adds reset() callout in identify docs


alanjcharles:
resolves #174

bsneed (approved):

PR #184, opened 2023-01-24
Fix to append rather than overwrite data in existing files

dannys42:
The existing code will result in malformed JSON being sent to the server because of partial events overwriting the start of existing files.  This fix will ensure we always append to a pre-existing file.

### DIFF
--- a/Sources/Segment/Utilities/OutputFileStream.swift
+++ b/Sources/Segment/Utilities/OutputFileStream.swift
@@ -53,6 +53,11 @@ internal class OutputFileStream {
         if fileHandle != nil { return }
         do {
             fileHandle = try FileHandle(forWritingTo: fileURL)
+            if #available(iOS 13.4, *) {
+                _ = try? fileHandle?.seekToEnd()
+            } else {
+                try? fileHandle?.seek(toOffset: .max)
+            }
         } catch {
             throw OutputStreamError.unableToOpen(fileURL.path)
         }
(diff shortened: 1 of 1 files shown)


Sources/Segment/Utilities/OutputFileStream.swift
```diff
@@ -53,6 +53,11 @@ internal class OutputFileStream {
         if fileHandle != nil { return }
         do {
             fileHandle = try FileHandle(forWritingTo: fileURL)
+            if #available(iOS 13.4, *) {
+                _ = try? fileHandle?.seekToEnd()
+            } else {
+                try? fileHandle?.seek(toOffset: .max)
+            }
```
dmcgloin: Likely could improve with code for all supported platforms

dmcgloin (changes requested):

dmcgloin (approved):

dmcgloin (approved):

bsneed:
Thanks @dannys42 @dmcgloin!  I haven't seen this happen before (not saying it's not possible, just that I haven't seen it) in my testing or test suites.  It definitely can't hurt for sure though.  Thanks so much for the PR!

dmcgloin:
Thanks for merging, @bsneed. Any idea when a release is targeted with this change?

bsneed:
Just released.  Thanks @dmcgloin !  https://github.com/segmentio/analytics-swift/releases/tag/1.3.5

PR #199, opened 2023-03-07
Null userId and traits update fixes

bsneed:
- Update objc identify method to accept nil userid and allow trait updates.
- Added test

### DIFF
--- a/Sources/Segment/ObjC/ObjCAnalytics.swift
+++ b/Sources/Segment/ObjC/ObjCAnalytics.swift
@@ -65,6 +65,8 @@ extension ObjCAnalytics {
     @objc(identify:traits:)
     public func identify(userId: String?, traits: [String: Any]?) {
         if let userId = userId {
+            // at first glance this looks like recursion.  It's actually calling
+            // into the swift version of this call where userId is NOT optional.
             analytics.identify(userId: userId, traits: traits)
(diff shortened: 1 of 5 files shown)


alanjcharles (approved):

Sources/Segment/ObjC/ObjCAnalytics.swift
```diff
@@ -63,8 +63,16 @@ extension ObjCAnalytics {
     ///   - traits: A dictionary of traits you know about the user. Things like: email, name, plan, etc.
     /// In the case when user logs out, make sure to call ``reset()`` to clear user's identity info.
     @objc(identify:traits:)
-    public func identify(userId: String, traits: [String: Any]?) {
-        analytics.identify(userId: userId, traits: traits)
+    public func identify(userId: String?, traits: [String: Any]?) {
+        if let userId = userId {
+            analytics.identify(userId: userId, traits: traits)
```
didiergarcia, line 70: This looks like a recursive infinite loop? 

On line 68: is userId not an Optional anymore? and is there identify() function that doesn't take an optional that this will go to instead?

bsneed: This is calling into the swift version of identify where userId is not optional.

didiergarcia (approved):

PR #219, opened 2023-04-24
Objective-C API improvements

bsneed:
- Make it such that ObjC can create / use enrichment plugins.
- Make Swift->ObjC mapping more 1 to 1.
- Documentation for SEGConfiguration and Configuration.
- Simplify event modification on the ObjC side.
- Add necessary tests; update existing.


Sources/Segment/Configuration.swift
```diff
@@ -45,60 +48,128 @@ public class Configuration {
...
+    /// ```
+    /// let defaults = Settings.load(resource: "mySegmentSettings.json")
+    /// let config = Configuration(writeKey: "1234").defaultSettings(defaults)
+    /// ```
+    ///
+    /// - Parameter settings: 
+    /// - Returns: The current Configuration.
     @discardableResult
-    func defaultSettings(_ settings: Settings) -> Configuration {
+    func defaultSettings(_ settings: Settings?) -> Configuration {
```
didiergarcia, line 115: I guess this means that values.defaultSettings can now be nil. Any downstream implications?

bsneed: Yup.  No downstream implications.  People weren't passing in nil before, so whatever they're passing now will continue.

Sources/Segment/ObjC/ObjCConfiguration.swift
```diff
@@ -13,6 +13,9 @@ import Foundation
 public class ObjCConfiguration: NSObject {
     internal var configuration: Configuration
     
+    /// Sets a reference to your application.  This can be useful in instances
+    /// where referring back to your application is necessary, such as within plugins
+    /// or async code.  The default value is `nil`.
     @objc
     public var application: Any? {
```
didiergarcia, line 20: Any chance this should be a WEAK reference?

Sources/Segment/ObjC/ObjCEvents.swift
```diff
@@ -0,0 +1,372 @@
...
+    case let e as TrackEvent:
+        return ObjCTrackEvent(event: e)
+    case let e as IdentifyEvent:
+        return ObjCIdentifyEvent(event: e)
+    case let e as ScreenEvent:
+        return ObjCScreenEvent(event: e)
+    case let e as GroupEvent:
+        return ObjCGroupEvent(event: e)
+    case let e as AliasEvent:
+        return ObjCAliasEvent(event: e)
```
didiergarcia, line 77: Do you have a page event that could go here?

didiergarcia (approved):

Sources/Segment/ObjC/ObjCConfiguration.swift
bsneed: Yeah, good call.

Sources/Segment/ObjC/ObjCEvents.swift
bsneed: Not yet.

Sources/Segment/ObjC/ObjCConfiguration.swift
bsneed: actually, not necessary.  it's a pass through and is already held weak.

PR #230, opened 2023-05-25
feat: add safely

alanjcharles:


Examples/apps/BasicExample/BasicExample/AppDelegate.swift
```diff
@@ -16,7 +16,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         // Override point for customization after application launch.
         
-        let configuration = Configuration(writeKey: "<WRITE KEY>")
+        let configuration = Configuration(writeKey: "7O86AjgXwDLHxwXapRDkSFvqRzLA5n1V")
```
didiergarcia: probably need to remove this.

Sources/Segment/Utilities/Storage.swift
```diff
@@ -54,7 +58,11 @@ internal class Storage: Subscriber {
             default:
                 if isBasicType(value: value) {
                     // we can write it like normal
-                    userDefaults?.set(value, forKey: key.rawValue)
+                    let setUserDefaults = self.userDefaults?.safelySetUserDefaults(valueToWrite: value, keyToWrite: key.rawValue)
```
didiergarcia: Should the name of this variable be something like "userDefaultsError" ?

Sources/Segment/Utilities/Storage.swift
```diff
@@ -54,7 +58,11 @@ internal class Storage: Subscriber {
...
+                    
+                    if let error = setUserDefaults {
+                        analytics?.reportInternalError(error)
```
didiergarcia, line 75: Does `analytics.reportInternalError(error) handle that type of error?

Sources/Segment/Utilities/Storage.swift
```diff
@@ -354,8 +362,34 @@ extension Storage {
...
+    struct Scenarios {
+        static let nullDefaultValues = SafeScenario(
+            description: "Guard against null getting set as a user default value",
+            implementor: "@alanjcharles")
+    }
+    
+    func safelySetUserDefaults( valueToWrite: Sendable, keyToWrite: String) -> Error? {
+        let context = UserContext(UserDefaults: UserDefaults(), valueToWrite: valueToWrite, keyToWrite: keyToWrite)
+        
+        let result = Safely.call(scenario: Scenarios.nullDefaultValues, context: context) { context in
```
didiergarcia: I think "result" should be named something like "callError".

Sources/Segment/Utilities/Storage.swift
alanjcharles: I think so because the type is the same I'm returning in `safelySetUserDefaults`. here's the method: 

```
    public func reportInternalError(_ error: Error, fatal: Bool = false) {
        let translatedError = Self.translate(error: error)
        configuration.values.errorHandler?(translatedError)
        Self.segmentLog(message: "An internal error occurred: \(translatedError)", kind: .error)
        if fatal {
            exceptionFailure("A critical error occurred: \(translatedError)")
        }
    }
   ```

Examples/apps/BasicExample/BasicExample/AppDelegate.swift
alanjcharles: yep lol every time i swear

Issue #229, opened 2023-05-18
App hanging when accessing userAgent value


everuribe:
**Describe the bug**
Since we started tracking app not responding instances, we have seen that many of our users experience app hanging of more than 10 seconds when `AppleUtils.userAgent` is accessed, specifically
```
WKWebView().value(forKey: "userAgent") as? String
```

This happens when analytics is configured (app launch). 

**To Reproduce**
Not consistently reproducible. 

**Expected behavior**
Perhaps this value could be set asynchronously, not in main thread, once it is available (unless this value must be accessed on main...). Or the option to not include this property could be provided. 

**Screenshots**
<img width="1496" alt="image" src="https://github.com/segmentio/analytics-swift/assets/47759786/108d3ec1-716e-4a48-b50d-d936ab849bc0">

**Platform (please complete the following information):**
 - Library Version in use: 1.4.0
 - Platform being tested: macOS

bsneed:
That's really weird, thanks for pointing it out!  I'm sure we can lazy fill it somehow.  I'll swing back to this with a PR in a bit.

bsneed:
Fixed!  A couple more issues to look at and then a release.

bsneed:
Fixed in 1.4.6

PR #236, opened 2023-06-26
Fixes for userAgent delay.  Retrieved async now.

bsneed:
- Made userAgent retrieval async.
- Updated context test to prime it.
- Moved it to the dynamic section of context collection.
- Fixes issue #229

### DIFF
--- a/Sources/Segment/Plugins/Platforms/Vendors/AppleUtils.swift
+++ b/Sources/Segment/Plugins/Platforms/Vendors/AppleUtils.swift
@@ -19,6 +19,7 @@ import WebKit
 
 internal class iOSVendorSystem: VendorSystem {
     private let device = UIDevice.current
+    @Atomic private static var asyncUserAgent: String? = nil
     
     override var manufacturer: String {
         return "Apple"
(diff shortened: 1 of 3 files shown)


alanjcharles (approved):

Sources/Segment/Plugins/Platforms/Vendors/AppleUtils.swift
```diff
@@ -65,17 +66,15 @@ internal class iOSVendorSystem: VendorSystem {
...
-        if Thread.isMainThread {
-            userAgent = WKWebView().value(forKey: "userAgent") as? String
-        } else {
-            DispatchQueue.main.sync {
-              userAgent = WKWebView().value(forKey: "userAgent") as? String
+        // BKS: It was discovered that on some platforms there can be a delay in retrieval.
+        // It has to be fetched on the main thread, so we've spun it off
+        // async and cache it when it comes back.
+        if Self.asyncUserAgent == nil {
+            DispatchQueue.main.async {
```
didiergarcia, line 73: Hey any chance you should re-check the condition `asyncUserAgent == nil` once you're in there?  This reminds me of the patterns that would require double-check locking, no?

Sources/Segment/Plugins/Platforms/Vendors/AppleUtils.swift
```diff
@@ -238,16 +238,15 @@ internal class MacOSVendorSystem: VendorSystem {
...
+                Self.asyncUserAgent = WKWebView().value(forKey: "userAgent") as? String
```
didiergarcia, line 246: Same question in here: Should we check the `asyncUserAgent == nil` condition here again?

Sources/Segment/Plugins/Platforms/Vendors/AppleUtils.swift
bsneed: I don't mind if it gets set twice.  It's locked via the @Atomic marker on the var so no race condition on double-set.

Sources/Segment/Plugins/Platforms/Vendors/AppleUtils.swift
bsneed: I don't mind if it gets set twice. It's locked via the @atomic marker on the var so no race condition on double-set.

Tests/Segment-Tests/Analytics_Tests.swift
```diff
@@ -123,6 +123,12 @@ final class Analytics_Tests: XCTestCase {
         let outputReader = OutputReaderPlugin()
         analytics.add(plugin: outputReader)
         
+        // prime the pump for userAgent, since it's retrieved async.
+        let vendorSystem = VendorSystem.current
+        while vendorSystem.userAgent == nil {
+            RunLoop.main.run(until: Date.distantPast)
```
didiergarcia, line 130: Is this a tight loop? basically spinning until that variable is set? If so, does a something like a count down latch fit more here?

didiergarcia (approved):

Issue #243, opened 2023-07-17
`context.app.name` is empty if `CFBundleDisplayName` is not included. `CFBundleDisplayName` is not required by Apple unless localized


dmcgloin:
**Describe the bug**
Expectation is that `context.app.name` should have the value from `CFBundleName` instead of being empty since the `CFBundleDisplayName` is not required by Apple for non-localized apps.

There is an open design question about how the distinction between `CFBundleName` and `CFBundleDisplayName` should be handled in Segment. I assume that in most cases the name is actually not localized, so the distinction is not relevant. But some customers may expect a constant name for all localized variants when accessing their reporting/filtering.

In any case, it feels like falling back to `CFBundleName` if `CFBundleDisplayName` is not present, could make sense.

Relevant source: https://github.com/segmentio/analytics-swift/blob/c8fd5fdf59299f00b3e4303a1b12a6d88893bf56/Sources/Segment/Plugins/Context.swift#L76

**To Reproduce**
Run an app without `CFBundleDisplayName`.
Observe any event in Segment trace event debugger that `context.app.name` is empty.

**Expected behavior**
Expected name to be present with value from `CFBundleName`.

**Platform (please complete the following information):**
 - Library Version in use: 1.4.7
 - iOS
 - Integrations in use: Custom

**Additional context**
More discussion here: https://stackoverflow.com/questions/3465792/whats-the-difference-between-bundle-display-name-and-bundle-name-in-cocoa-a

PR #252, opened 2023-09-06
fix: add 'CFBundleName' to app context

alanjcharles:
- adds `CFBundleName` as fallback for `app.name` in context


alanjcharles:
resolves #243

bsneed (approved):

PR #256, opened 2023-10-04
Update Flush policies

niallzato:
Default visibility of init method will be `internal`, add `public` prefix.

Also add test for the interval policy.


alanjcharles (approved):
great work dawg!

PR #259, opened 2023-10-16
Make Segment.io flush asynchronous

### LINKED ISSUE #257: Flush running on main thread and may take a long time
**Describe the bug**
We recently have switched over to using your Segment Swift-based library (from your Objective-C library).
We have code in our app that checks whether the main thread is blocked for relatively long periods of time.
We have noticed that:
(a) flush operations take place on the main thread, and
(b) those flush operations can take significant amounts of time. E.g., > 3 seconds. 

**To Reproduce**
1. Turn off all networking on device. E.g., airplane mode.
2. Log numerous Segment events. 
3. Turn networking back on on device.
4. Observe that flush takes place on main thread and that main thread is blocked for a relatively long period of time.

**Expected behavior**
Given that the flush can take a relatively long period of time (roughly proportional to the number of events pending to be sent), I expect the flush to not take place on the main thread-- which will lock up the UI if the app is in the foreground.

**Screenshots**
N/A

 - Library Version in use: 1.4.7
 - Platform being tested: iOS

**Additional context**
Our app serves pilots flying aircraft and in this context the devices typically do not have a networking connection, but Segment events are logged during flights. When the pilot lands, and again has networking, Segment events get uploaded.

**Questions**

1. We are planning to work around this issue by using custom flush policies which directly call the flush operation (e.g., as in IntervalBasedFlushPolicy), but which wrap the flush call in use of a background thread. Do you see any problematic consequences of this strategy in terms of the broader Segment library and system? So far in initial testing it works with no problem.
 
2. Would you be open to a PR to change this? The simplest approach seems to me to [wrap the bulk of the flush code](https://github.com/segmentio/analytics-swift/blob/80711a259fcd26e3ef44eb7a9b92f4017579188d/Sources/Segment/Analytics.swift#L204-L208) in a block executing on a background thread. Example:

```
    public func flush() {
        // only flush if we're enabled.
        guard enabled == true else { return }
        
        DispatchQueue.global(qos: .background).async {
            apply { plugin in
                if let p = plugin as? EventPlugin {
                    p.flush()
                }
            }
        }
    }
```

bsneed:
- Make segment.io flush an asynchronous operation.
- Fixed naming convention for serial queue in segment destination.
- Fixes #257 

It's my suspicion that file IO is the slow-down when calling flush() from the main thread with a significant number of events present.  As such, in this fix async operation was done in the segment destination instead of the upper level analytics method to prevent unknowns from happening with device mode destinations.  Those device mode destinations (now or in the future) may not be as thread-safe, or have expectations about running on the main thread, etc.

A future-todo would be to make asynchronous file iO standard and remove usage of FileManager and the like.

### DIFF
--- a/Sources/Segment/Plugins/SegmentDestination.swift
+++ b/Sources/Segment/Plugins/SegmentDestination.swift
@@ -42,7 +42,8 @@ public class SegmentDestination: DestinationPlugin, Subscriber {
     
     internal var httpClient: HTTPClient?
     private var uploads = [UploadTaskInfo]()
-    private let uploadsQueue = DispatchQueue(label: "uploadsQueue.segment.com")
+    private let uploadsQueue = DispatchQueue(label: "com.segment.uploadsQueue")
+    private let flushQueue = DispatchQueue(label: "com.segment.flushQueue")
     private var storage: Storage?
     
     @Atomic internal var eventCount: Int = 0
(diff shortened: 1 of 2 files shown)


bsneed:
Some tests need to be fixed due to the new async nature of flush and tests depending on the previous sync.

Sources/Segment/Plugins/SegmentDestination.swift
```diff
@@ -113,47 +114,50 @@ public class SegmentDestination: DestinationPlugin, Subscriber {
     }
     
     public func flush() {
```
cprince-foreflight, line 116: It seems like this could be simplified to:
```
public func flush() {
    flushQueue.async { [weak self] in
        self?._flush()
    }
}

// Existing flush (no need for this comment-- just making the point)
private func _flush() {
  // existing code...
}
```

bsneed:
@cprince-foreflight can you give the branch a try and let me know the results?

cprince-foreflight:
> @cprince-foreflight can you give the branch a try and let me know the results?

Just tried a smoke test of this branch with our app and am still seeing the expected Segment events in the Segment test debugger. So, no issues apparent. (Though it takes longer-- with builds to our beta testers-- to get real tests of the issue that brought this about. I.e., tests with the app in flight).

Any ideas of when (a) this will be merged and (b) when a tagged release with this will happen?

Thanks!

bsneed:
@cprince-foreflight Some version of this likely sometime next week.  We're having an internal dialogue about whether we want to make sync vs. async a choice at the API level for flush.  There's scenarios that demand one or the other and the lib wasn't structured to take this into account.

bsneed:
Closing this in favor of #269

PR #263, opened 2023-10-26
Add new fields to settings object decode.

bsneed:
- Added metrics to settings object decode.
- Added consentSettings to settings object decode.

### DIFF
--- a/Sources/Segment/Settings.swift
+++ b/Sources/Segment/Settings.swift
@@ -12,6 +12,8 @@ public struct Settings: Codable {
     public var plan: JSON? = nil
     public var edgeFunction: JSON? = nil
     public var middlewareSettings: JSON? = nil
+    public var metrics: JSON? = nil
+    public var consentSettings: JSON? = nil
 
     public init(writeKey: String, apiHost: String) {
         integrations = try! JSON([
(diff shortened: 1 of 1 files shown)


Sources/Segment/Settings.swift
```diff
@@ -37,6 +39,8 @@ public struct Settings: Codable {
         self.plan = try? values.decode(JSON.self, forKey: CodingKeys.plan)
         self.edgeFunction = try? values.decode(JSON.self, forKey: CodingKeys.edgeFunction)
         self.middlewareSettings = try? values.decode(JSON.self, forKey: CodingKeys.middlewareSettings)
+        self.metrics = try? values.decode(JSON.self, forKey: CodingKeys.metrics)
+        self.consentSettings = try? values.decode(JSON.self, forKey: CodingKeys.consentSettings)
```
didiergarcia, line 43: Where is this `CodingKeys.consentSettings` defined?

didiergarcia (approved):

Sources/Segment/Settings.swift
bsneed: down below as 

```
    enum CodingKeys: String, CodingKey {
        case integrations
        case plan
        case edgeFunction
        case middlewareSettings
        case metrics
        case consentSettings
    }
 ```

PR #268, opened 2023-11-15
Allow configuration of custom userAgent

hbmartin:
We've encountered a need to configure a custom user agent for app. This PR enables setting that with a new field in `Configuration`


bsneed:
Hi @hbmartin, thanks for the contribution.  I think something like this is better done as a user-defined plugin.  Something alone the lines of this in it's simplest form.

```swift
analytics.add(enrichment: { event in
    event.context["userAgent"] = "myUserAgentString"
})
```

What do you think?  I'm not opposed to your change, but it adds core functionality that the plugin system is meant to address.

hbmartin:
@bsneed If I understand correctly, the big difference between this PR and the plugin approach is that the async work to get the device user agent would still happen and only later be overwritten by the plugin value. We would like to avoid that work entirely when setting a custom UA. (If I'm incorrect then happy to use the plugin!)

bsneed:
Ah, that's a great point.  Merge underway.  I've got another release in the works and it'll go out with that.

PR #269, opened 2023-11-16
Added operating mode config option & behavior.

bsneed:
- Added operatingMode confi option & logic
- Changed unit-test settings setup to respect operating mode.
- Made some startup plugins not auto-load when in server mode.
- Fixed logging output to be correct.
- Updated testPurgeStorage to reflect new flush behavior.
- Added test for server operation mode.

### DIFF
--- a/Sources/Segment/Configuration.swift
+++ b/Sources/Segment/Configuration.swift
@@ -13,10 +13,12 @@ import FoundationNetworking
 // MARK: - Operating Mode
 /// Specifies the operating mode/context
 public enum OperatingMode {
-    /// Running as an executable on a server
-    case server
-    /// Running as a long-lived application
-    case client
+    /// The operation of the Analytics client are synchronous.
+    case synchronous
+    /// The operation of the Analytics client are asynchronous.
+    case asynchronous
+    
+    static internal let defaultQueue = DispatchQueue(label: "com.segment.operatingModeQueue", qos: .utility)
 }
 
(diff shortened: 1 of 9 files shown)


Sources/Segment/Configuration.swift
```diff
@@ -10,6 +10,15 @@ import Foundation
 import FoundationNetworking
 #endif
 
+// MARK: - Operating Mode
+/// Specifies the operating mode/context
+public enum OperatingMode {
+    /// Running as an executable on a server
```
didiergarcia: Callout to project README.md to list differences.

didiergarcia (approved):

alanjcharles (approved):

Sources/Segment/Configuration.swift
```diff
@@ -10,6 +10,15 @@ import Foundation
...
+    case server
+    /// Running as a long-lived application
```
oscb: One could argue a server is longer-lived than an app! stateless might be a better word.

The other thing I'm thinking of is, technically you might want to do async in a server too. At least in other languages I have used for server side.

bsneed: I definitely see your point.  Would better enum values be sync/async? cc @didiergarcia

oscb: I think async/sync make it clearer what to expect when you switch!

Issue #284, opened 2023-12-14
Nit: Allow date formatting to be configured by client?


cprince-foreflight:
**Is your feature request related to a problem? Please describe.**
This is more of a nit than anything else, but it would be nice. Our app currently does date formatting using the `ISO8601DateFormatter()` configured with `formatter.formatOptions.insert(.withFractionalSeconds)`. In this way, we get a string date with a format like:
"2023-10-04T21:41:12.000Z"

The default date formatter in Segment currently formats dates slightly differently. For example:
"2023-12-13T23:26:37.091+0000"

**Describe the solution you'd like**
Could date formatting be configurable by the calling client? E.g., through supplying a `DateEncodingStrategy` for the JSONEncoder and `DateDecodingStrategy` for the JSONDecoder?

bsneed:
Hi @cprince-foreflight, yeah, definitely, it's a possibility!  Though, we don't want to go that route probably as we need consistent dates across the board.  Last I looked, we were getting the 000Z version you show but it has been a minute since I checked.  Segment internally needs 8601 dates universally, so the options provided by JSONDecoder are more than would be desirable.  If you have an idea and PR, I can probably get it merged pretty quick if it follows the 000Z format.  Otherwise, it'll likely be post holidays.

cprince-foreflight:
Looking at this more this morning, I'm realizing that the date formatting difference I'm finding has to do with using the Codable interface versus the non-Codable interface in Segment 1.5.0.

[_Codable interface_](https://github.com/segmentio/analytics-swift/blob/266c374bd9c55a3101e0b5f04a9f8cab4eee747a/Sources/Segment/Events.swift#L17)
Default date formatting:
"2023-12-13T23:26:37.091+0000"

[_non-Codable interface_](https://github.com/segmentio/analytics-swift/blob/266c374bd9c55a3101e0b5f04a9f8cab4eee747a/Sources/Segment/Events.swift#L145)
Default date formatting:
"2023-12-14T16:03:14.300Z"

Looking for the underlying difference in the code now.

cprince-foreflight:
I see it now. Different date formatters are being used:

_Codable_
[Code reference](https://github.com/segmentio/analytics-swift/blob/266c374bd9c55a3101e0b5f04a9f8cab4eee747a/Sources/Segment/Utilities/JSON.swift#L22)
```
extension DateFormatter {
    static let iso8601: DateFormatter = {
        let formatter = DateFormatter()
[...]

PR #285, opened 2023-12-14
Change to using a single date formatter and update unit tests.

cprince-foreflight:
https://github.com/segmentio/analytics-swift/issues/284

As described in the ticket above, the two date formatters in the library didn't match up in the strings they produce. I've consolidated to a single date formatter using the string format like:
"2023-12-14T16:03:14.300Z"

The format used by the other date formatter, that I've removed, was: 
"2023-12-13T23:26:37.091+0000"

_Testing_
* Changed a unit test to support this single date formatter.
* Did some initial testing in our own app by bringing the changed analytics-swift package in and looking at Segment debugger output for dates with Codable and non-Codable interfaces. Both now are the same in my tests.

### DIFF
--- a/Sources/Segment/Utilities/JSON.swift
+++ b/Sources/Segment/Utilities/JSON.swift
@@ -9,9 +9,23 @@ import Foundation
 import JSONSafeEncoder
 
 extension JSONDecoder {
+    enum JSONDecodingError: Error {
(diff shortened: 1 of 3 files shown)


bsneed (approved):

bsneed:
Released as 1.5.1

PR #294, opened 2024-02-12
Multiple instance write key

alanjcharles:
-add check in `Analytics` to prevent multiple instances with the same `write-key` from being initialized at the same time


guiihlopes:
@alanjcharles is it possible to have multiple instances with differents write keys?

alanjcharles:
@guiihlopes  yep, definitely. See the example here: https://github.com/segmentio/analytics-swift/blob/main/Examples/tasks/MultiInstance.swift

PR #302, opened 2024-02-28
Allow Codable compliant values in json's init(any).

### LINKED ISSUE #301: JSON encoding: lots of Codable-conforming types that crash in JSON encoding
**Describe the bug**
There are many swift data types that crash in the library's JSON handling that (I believe) shouldn't. And since the type that the JSON object accepts in its initializer is `[String: Any]`, the compiler can't catch any of these issues before runtime. This makes working with the library difficult (since you have to know which types are accepted and manually convert everything else to those types) and fragile (since you need to manually test each tracking call to ensure it doesn't crash the JSON logic).

**To Reproduce**
1. Do `segment.track(name: "Some Event Name", properties: ["key": value])` where the value is any of the following types (there are probably more, but these are the ones I've run into so far):
    - UUID
    - enum that conforms to Codable (e.g. `enum MyEnum: String, Codable { case a, b, c }`)
2. See the following error:
```
Segment/Utils.swift:54: Fatal error: A critical error occurred: jsonUnableToDeserialize(Segment.JSON.JSONError.nonJSONType(type: ...))
```

**Expected behavior**
Ideally, types that conform to `Encodable` and that can be properly serialized using a regular `JSONEncoder` should not crash in the library's handling of JSON

 - Library Version in use: 1.5.3
 - Platform being tested: iOS

bsneed:
- Allow codable compliant types in JSON's init(any).
- Added test
- Fixes #301

### DIFF
--- a/Sources/Segment/Utilities/JSON.swift
+++ b/Sources/Segment/Utilities/JSON.swift
@@ -76,7 +76,7 @@ public enum JSON: Equatable {
     public init<T: Codable>(with value: T) throws {
         let encoder = JSONSafeEncoder.default
         let json = try encoder.encode(value)
-        let output = try JSONSerialization.jsonObject(with: json)
+        let output = try JSONSerialization.jsonObject(with: json, options: .fragmentsAllowed)
         try self.init(output)
     }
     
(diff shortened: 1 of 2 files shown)


bsneed:
Test is still a WIP.

alanjcharles (approved):

PR #310, opened 2024-03-18
[Fix] Missing Privacy Manifest in Bundle

FelixHerrmann:
I have noticed that the privacy manifest that is included in this package is not part of the package's bundle and therefore cannot be accessed from Xcode and App Store Connect.

This PR moves the PrivacyInfo.xcprivacy to a dedicated `Resources` directory which is specified as resource in the package manifest.


miquelalvarado:
@FelixHerrmann Thank you for this! I was about to do the same thing. @bsneed Do you think it could be reviewed anytime soon? 🙏

bsneed (approved):

bsneed:
Thanks @FelixHerrmann!  Nice find, will get this released shortly.

PR #311, opened 2024-03-18
[Fix] Missing Privacy Tracking and API-Types

FelixHerrmann:
In addition to #310 I have noticed some missing informations in the privacy manifest:
- **NSPrivacyTracking**: A Boolean that indicates whether your app or third-party SDK uses data for tracking as defined under the App Tracking Transparency framework. For more information, see [User Privacy and Data Use](https://developer.apple.com/app-store/user-privacy-and-data-use/).
- **NSPrivacyAccessedAPITypes**: An array of dictionaries that describe the API types your app or third-party SDK accesses that have been designated as APIs that require reasons to access. For information on the keys and values to use in the dictionaries, see [Describing use of required reason API](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api).

Source: https://developer.apple.com/documentation/bundleresources/privacy_manifest_files#4284009

---

This PR add's those missing informations with the following considerations:
- despite the [IDFACollection plugin example](https://github.com/segmentio/analytics-swift/blob/main/Examples/other_plugins/IDFACollection.swift) the package itself is not doing any tracking
- the UserDefaults could potentially be used in not only Apps but AppGroups due to the use of suites, e.g. [here](https://github.com/segmentio/analytics-swift/blob/a59b6fe4b4b89e859a1ee02fee6872a4ae6032f7/Sources/Segment/Utilities/Storage/Storage.swift#L25)

Please double check all of that and confirm or adjust; all available UserDefault options are listed [here](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278401)!

---

~This will conflict with #310 but should be easy to resolve.~ ✅


bsneed (approved):
Looks good; I'll make some additional changes via a separate PR.  Thanks for updating this!

vlvs:
How come `NSPrivacyTracking` is set to `false` even though there is a list of `NSPrivacyTrackingDomains`? 🤔

PR #314, opened 2024-03-20
fix: compare default `apiHost` with `apiHost` configuration instead of `writeKey`

pietropizzi:
I just found this bug that when I override the `apiHost` through the configuration it is set back to the default host once the settings are fetched. Event though the settings don't have an overridden host.

That is because you compare the `host` against `configuration.values.writeKey` instead of `configuration.values.apiHost`.

This fixes that.


pietropizzi:
@alanjcharles @bsneed tagging you as most recent contributors. We are trying to migrate to analytics-swift from the legacy objc version and this is something we would need fixed. Please have a look.

bsneed (approved):

bsneed:
Thanks for finding that @pietropizzi !  We'll get this merged shortly.

PR #315, opened 2024-03-22
Updated privacy info

bsneed:
- Correct privacy domains to the proper format.
- Removed location privacy info, since location is NOT collected.


bsneed:
cc @FelixHerrmann

PR #318, opened 2024-03-27
Remove privacy domains

bsneed:
- Fixes #315 

From: https://developer.apple.com/app-store/app-privacy-details/#user-tracking

**“Tracking” refers to linking data collected from your app about a particular end-user or device, such as a user ID, device ID, or profile, with Third-Party Data for targeted advertising or advertising measurement purposes, or sharing data collected from your app about a particular end-user or device with a data broker.**

We do not collect data for advertising or measurement purposes, and we are not a data broker.  These decisions are up to the user and marked as such in their own apps.

### DIFF
--- a/Sources/Segment/Resources/PrivacyInfo.xcprivacy
+++ b/Sources/Segment/Resources/PrivacyInfo.xcprivacy
@@ -4,11 +4,6 @@
 <dict>
 	<key>NSPrivacyTracking</key>
 	<false/>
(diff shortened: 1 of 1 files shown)


slashignore:
You should update the documentation around the tracking domains and the tracking transparency framework https://segment.com/docs/connections/sources/catalog/libraries/mobile/apple/ios-17/

PR #319, opened 2024-03-27
docs: add apple doc for track() apis

jamesonwilliams:
Apple docs were missing from two of the `track()` APIs, and for another, the docs had been copy-pasted from `identify()` without being updated. This updates all three APIs to something simple and functional.


alanjcharles:
thanks @jamesonwilliams !

PR #363, opened 2024-09-18
No longer prompt user for local network access on app launch on macOS Sequoia

manavgabhawala:
Calling `ProcessInfo.processInfo.hostName` causes a new TCC prompt on macOS Sequoia for the user requesting the app be granted local network access. This code path is hit pretty often, typically during app launch. However, nothing in this framework inherently requires local network access, thus by changing the property to `ProcessInfo.processInfo.userName` instead, any app linking this framework will no longer be forced to prompt for local network access.

### DIFF
--- a/Sources/Segment/Plugins/Platforms/Vendors/AppleUtils.swift
+++ b/Sources/Segment/Plugins/Platforms/Vendors/AppleUtils.swift
@@ -201,10 +201,6 @@ internal class MacOSVendorSystem: VendorSystem {
         return deviceModel()
     }
     
-    override var name: String {
-        return device.hostName
-    }
-    
     override var identifierForVendor: String? {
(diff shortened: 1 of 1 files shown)


bsneed:
Thanks @manavgabhawala!  Unfortunately, we can't use "username" there as it would compromise part of the login for that machine.  It's probably best to remove that override and let the base classes "unknown" value come through.  I quick glance didn't show any alternatives that'd produce the same value without the undesirable effect.

If you can do that, I'll merge this PR.  Thanks so much!

manavgabhawala:
@bsneed updated the PR using your recommendation! Thanks

bsneed (approved):

bsneed:
thanks so much @manavgabhawala !

PR #366, opened 2024-10-04
feat: add AnalyticsError type for setting fetch failed.

didiergarcia:
Adding AnalyticsError.settingsFetchFailed(Error) type.

Wrap all the errors that occur while fetching settings using the new type.

### DIFF
--- a/Sources/Segment/Utilities/Networking/HTTPClient.swift
+++ b/Sources/Segment/Utilities/Networking/HTTPClient.swift
@@ -63,7 +63,7 @@ public class HTTPClient {
 
         let dataTask = session.uploadTask(with: urlRequest, fromFile: batch) { [weak self] (data, response, error) in
             guard let self else { return }
-            handleResponse(data: data, response: response, error: error, completion: completion)
+            handleResponse(data: data, response: response, error: error, urlRequest: urlRequest, completion: completion)
         }
 
         dataTask.resume()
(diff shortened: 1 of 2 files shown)


alanjcharles (approved):

Sources/Segment/Utilities/Networking/HTTPClient.swift
```diff
@@ -63,7 +63,7 @@ public class HTTPClient {
 
         let dataTask = session.uploadTask(with: urlRequest, fromFile: batch) { [weak self] (data, response, error) in
             guard let self else { return }
-            handleResponse(data: data, response: response, error: error, completion: completion)
+            handleResponse(data: data, response: response, error: error, urlRequest: urlRequest, completion: completion)
```
bsneed: if we just need the url, lets just pass the url instead of the whole request.

bsneed (approved):

alanjcharles (approved):
lgtm!

alanjcharles:
although I am not sure what's going on with the linux test CC @bsneed

bsneed:
looks like that disk image is now gone; can explore fixing it later.

Sources/Segment/Utilities/Networking/HTTPClient.swift
didiergarcia: Done

PR #369, opened 2024-10-15
Telemetry module for Swift

MichaelGHSeg:


Sources/Segment/Utilities/Telemetry.swift
```diff
@@ -0,0 +1,278 @@
...
+    private var queueBytes = 0
+    private var queueSizeExceeded = false
+    private var seenErrors = [String: Int]()
+    public var started = false
+    private var rateLimitEndTime: TimeInterval = 0
+    private var telemetryQueue = DispatchQueue(label: "telemetryQueue")
+    private var telemetryTimer: Timer?
+
+    func start() {
+        guard enable, !started, sampleRate != 0.0 else { return }
```
didiergarcia: Maybe `sampleRate > 0.0 && sampleRate <= 1.0` ?

Sources/Segment/Utilities/Telemetry.swift
```diff
@@ -0,0 +1,278 @@
...
+
+        return request
+    }
+
+    private func queueHasSpace() -> Bool {
+        var over = false
+        telemetryQueue.sync {
+            over = queue.count < maxQueueSize
+        }
+        return over
```
didiergarcia: I think value determined is if you are "under" max queue size. Should that variable be changed to `under` ?

didiergarcia (approved):
Looks good overall.

I would add some better protection to the sample rate by given it a bounds like > 0.0 and <= 1.0 if that's the intension.

Sources/Segment/Utilities/Telemetry.swift
MichaelGHSeg: That would make more sense.  I like the guard clause but the inverted logic got me a bit.

PR #383, opened 2025-03-11
Don't output traits if they are empty.

bsneed:
- Added OmitIfNull property wrapper to avoid outputting keys with values that are null.
- Modified context traits to only output if non-null.
- Added test.


didiergarcia (approved):

bsneed:
Closing this PR as it's unnecessary.  Empty traits are being added server-side, not the library.

PR #385, opened 2025-04-03
feat: Allow for more granular tracked lifecycle events to track

delannoyk:
Hi Segment team 👋🏻,

We found ourselves not wanting to track all the lifecycle events in our app. As far as I can tell there was no way to filter them so I implemented this as an OptionSet instead of a boolean value.

Let me know if this is acceptable and/or if you have any questions or suggestions.

Thanks.
Kevin

### DIFF
--- a/Sources/Segment/ObjC/ObjCConfiguration.swift
+++ b/Sources/Segment/ObjC/ObjCConfiguration.swift
@@ -29,12 +29,12 @@ public class ObjCConfiguration: NSObject {
 
     /// Opt-in/out of tracking lifecycle events.  The default value is `false`.
     @objc
-    public var trackApplicationLifecycleEvents: Bool {
+    public var trackApplicationLifecycleEvents: TrackedLifecycleEvent {
         get {
-            return configuration.values.trackApplicationLifecycleEvents
+            return configuration.values.trackedApplicationLifecycleEvents
         }
         set(value) {
-            configuration.trackApplicationLifecycleEvents(value)
+            configuration.setTrackedApplicationLifecycleEvents(value)
         }
     }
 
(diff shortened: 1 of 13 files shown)


bsneed:
I can't merge this as it breaks the API.  The optimal way to do this is disable lifecycle events, and track on the ones you do, either in your code or via a plugin similar to the lifecycle plugin in this codebase.

delannoyk:
@bsneed I made sure the API wasn't broken when implementing this, would you mind explaining a bit more? I don't mind refactoring it...
If that's a no for other reasons, that's fine too, I just thought I'd ask for more precision on breaking the API?

bsneed:
@delannoyk you totally did!  I'm so sorry I missed that.  I shoulda dug deeper, so thanks for inquiring about it.  I'll reopen this and give it a whirl in just a bit.

delannoyk:
@bsneed appreciate it, but also no pressure! I was just wondering. Thanks for considering it!

Sources/Segment/ObjC/ObjCConfiguration.swift
```diff
@@ -29,12 +29,12 @@ public class ObjCConfiguration: NSObject {
 
     /// Opt-in/out of tracking lifecycle events.  The default value is `false`.
     @objc
-    public var trackApplicationLifecycleEvents: Bool {
+    public var trackApplicationLifecycleEvents: TrackedLifecycleEvent {
```
bsneed, line 32: I think this needs to stay as a bool to preserve compatibility.  can change it on my end though.

bsneed (commented):
looks good.  I'll need to verify a recent add of appWillTerminate and can get this merged and do those fix-ups.  Thanks @delannoyk !

bsneed:
Merged with fix-ups in #396

PR #386, opened 2025-04-09
Remove useless Codable requirement

arnauddorgans:
The PR removes useless Codable requirements

### DIFF
--- a/Examples/other_plugins/NotificationTracking.swift
+++ b/Examples/other_plugins/NotificationTracking.swift
@@ -42,7 +42,7 @@ class NotificationTracking: Plugin {
     var type: PluginType = .utility
     weak var analytics: Analytics?
 
-    func trackNotification(_ properties: [String: Codable], fromLaunch launch: Bool) {
+    func trackNotification(_ properties: [String: Encodable], fromLaunch launch: Bool) {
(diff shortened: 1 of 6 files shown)


bsneed:
can you elaborate on why you needed to make this change, or was it purely for correctness?

arnauddorgans:
@bsneed  it forces us to write useless Decodable conformances on objects that are not decodable
e.g.

```swift
extension AnalyticsClient {
    struct Context: Codable {
        let value: Any
        let encode: (inout SingleValueEncodingContainer) throws -> Void

        init<T: Encodable>(_ value: T) {
            self.value = value
            self.encode = { try $0.encode(value) }
        }

        init(from decoder: any Decoder) throws {
            // Waiting for https://github.com/segmentio/analytics-swift/pull/386 to me merged
            throw DecodingError.dataCorrupted(.init(
                codingPath: decoder.codingPath,
                debugDescription: "Decodable is not used"
            ))
        }

        func encode(to encoder: any Encoder) throws {
            var container = encoder.singleValueContainer()
            try encode(&container)
        }
    }
}
```

bsneed:
Thanks for the explanation @arnauddorgans

PR #394, opened 2025-07-04
fix: extract NSURL from launch options

granbestiapop:
## Description
When extracting the url from launch options we always receive empty string. Using the debugger the spec changed to NSURL instead of String.
With this change, two ways are considered, String as it and NSURL 
Screenshot 2025-07-04 at 9 41 55 AM


juandelgadojusto (approved):

bsneed:
Thanks @juandelgadojusto !

PR #396, opened 2025-07-13
Performed fixup for url options, lifecycle updates, objc compatibility

bsneed:
- Fixed merge conflict for url options.
- Verified willTerminate's recent wire-up was still appropriate.
- Moved trackApplicationLifecycleEvents back to being a bool.
- Fixed relevant examples

### DIFF
--- a/Examples/apps/BasicExample/BasicExample/AppDelegate.swift
+++ b/Examples/apps/BasicExample/BasicExample/AppDelegate.swift
@@ -17,7 +17,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // Override point for customization after application launch.
         
         let configuration = Configuration(writeKey: "WRITE KEY")
-            .trackApplicationLifecycleEvents(true)
+            .setTrackedApplicationLifecycleEvents(.all)
             .flushInterval(10)
             .flushAt(2)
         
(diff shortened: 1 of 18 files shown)


Examples/apps/BasicExample/BasicExample/AppDelegate.swift
```diff
@@ -17,7 +17,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // Override point for customization after application launch.
         
         let configuration = Configuration(writeKey: "WRITE KEY")
-            .trackApplicationLifecycleEvents(true)
+            .setTrackedApplicationLifecycleEvents(.all)
```
didiergarcia, line 20: This is "Moved trackApplicationLifecycleEvents back to being a bool." ?

Tests/Segment-Tests/Support/TestUtilities.swift
```diff
@@ -178,7 +178,7 @@ extension XCTestCase {
     
     func waitUntilFinished(analytics: Analytics?, file: StaticString = #filePath, line: UInt = #line) {
         addTeardownBlock { [weak analytics] in
-            let instance = try await waitForTaskCompletion(withTimeoutInSeconds: 3) {
+            let instance = try await waitForTaskCompletion(withTimeoutInSeconds: 30) {
```
didiergarcia, line 181: I guess we're assuming the test will be slower? Should we use a tighter limit?

didiergarcia (approved):

Examples/apps/BasicExample/BasicExample/AppDelegate.swift
bsneed: No, the objc interface was moved back to being a bool for compatibility.

Tests/Segment-Tests/Support/TestUtilities.swift
bsneed: The iPhone ci runner is all over the map speed wise; getting a lot of false positives for this test failing.

PR #398, opened 2025-07-30
Create SegmentTestApp

Hewadstanikzay:
This pull request adds a basic SwiftUI application that integrates Segment Analytics using the Segment Swift SDK. It includes a configuration with writeKey and a simple ContentView.  
This will help verify if Segment events are correctly sent from the iOS app.

### DIFF
--- a/SegmentTestApp
+++ b/SegmentTestApp
@@ -1,0 +1,14 @@
+import Segment
+
+@main
+struct SegmentTestApp: App {
+    init() {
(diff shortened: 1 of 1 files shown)


bsneed:
There's already a swiftUI example app here: https://github.com/segmentio/analytics-swift/tree/main/Examples/apps/SegmentSwiftUIExample

Is there something missing from that one?

PR #399, opened 2025-08-01
Use appropriate storage location + migrate events

bsneed:
- Use applicationSupportDirectory for storage instead of documents/cache.
- Migrate any existing event batches.
- Added test to verify migration.

### DIFF
--- a/Sources/Segment/Utilities/Utils.swift
+++ b/Sources/Segment/Utilities/Utils.swift
@@ -98,21 +98,22 @@ private func migrateFromOldLocations(writeKey: String, to newLocation: URL) {
     // If segment dir already exists in app support, we're done
     guard !fm.fileExists(atPath: newSegmentDir.path) else { return }
     
-    // Look for old segment directories to move
-    let oldLocations = [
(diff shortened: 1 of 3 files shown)


Sources/Segment/Utilities/Utils.swift
```diff
@@ -75,17 +75,44 @@ extension Optional: Flattenable {
...
+        fm.urls(for: .cachesDirectory, in: .userDomainMask).first?.appendingPathComponent("segment")
+    ].compactMap { $0 }
+    
+    for oldSegmentDir in oldLocations {
+        guard fm.fileExists(atPath: oldSegmentDir.path) else { continue }
+        
+        do {
+            try fm.moveItem(at: oldSegmentDir, to: newSegmentDir)
+            Analytics.segmentLog(message: "Migrated analytics data from \(oldSegmentDir.path)", kind: .debug)
+            return // Success!
```
didiergarcia: Is this returning out of the for-loop and then function? Should this be a break?

didiergarcia (approved):

PR #400, opened 2025-09-16
Remove objc compatibility from TrackedLifecycleEvent since it's not exposed to objc.

bsneed:
- Removed @objc from TrackedLifeCycleEvent since it's not exposed to objc.


MichaelGHSeg (approved):

bsneed:
will fix ci separately.

PR #401, opened 2025-09-19
Correct time calculations

dfed:
All but one of the extensions on `TimeInterval` in `QueueTimer.swift` yielded incorrect values.

### DIFF
--- a/Sources/Segment/Utilities/QueueTimer.swift
+++ b/Sources/Segment/Utilities/QueueTimer.swift
@@ -73,7 +73,7 @@ internal class QueueTimer {
 
 extension TimeInterval {
     static func milliseconds(_ value: Int) -> TimeInterval {
-        return TimeInterval(value / 1000)
+        return TimeInterval(value) / 1000
     }
     
     static func seconds(_ value: Int) -> TimeInterval {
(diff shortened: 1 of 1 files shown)


Sources/Segment/Utilities/QueueTimer.swift
```diff
@@ -73,18 +73,18 @@ internal class QueueTimer {
 
 extension TimeInterval {
     static func milliseconds(_ value: Int) -> TimeInterval {
-        return TimeInterval(value / 1000)
+        return TimeInterval(value) / 1000
```
dfed, line 76: This calculation had an integer-division problem. My change makes it so that we're dividing `Double` types rather than `Int` types, and therefore are no longer truncating.

Sources/Segment/Utilities/QueueTimer.swift
```diff
@@ -73,18 +73,18 @@ internal class QueueTimer {
...
     }
     
     static func seconds(_ value: Int) -> TimeInterval {
         return TimeInterval(value)
     }
     
     static func hours(_ value: Int) -> TimeInterval {
-        return TimeInterval(60 * value)
+        return TimeInterval(60 * 60 * value)
```
dfed, line 84: a `TimeInterval` represents a second. There are 60 seconds in a minute, and 60 seconds in an hour.

Sources/Segment/Utilities/QueueTimer.swift
```diff
@@ -73,18 +73,18 @@ internal class QueueTimer {
...
     }
     
     static func days(_ value: Int) -> TimeInterval {
-        return TimeInterval((60 * value) * 24)
+        return TimeInterval((60 * 60 * value) * 24)
```
dfed, line 88: Same as above.

bsneed (approved):
Oh man, how embarrassing.   Thanks for finding it and fixing!!

PR #419, opened 2026-02-27
Fix race condition causing batch payload corruption

didiergarcia:
## Problem

Customer reported intermittent batch payload corruption in analytics-swift v1.9.1 where events were appended outside the batch array structure, causing silent data loss.

### Symptoms
- Batch array closes prematurely after first event
- Subsequent events appended as raw JSON after closing bracket  
- Server returns 200 OK but only processes first event (rest silently dropped)
- No error reported to application

### Example Corrupted Payload
```json
{
  "batch": [
    {event1}
  ],
  "sentAt": "...",
  "writeKey": "..."
}
{event2},
{event3},
{event4}
```

## Root Cause

Race condition in `TransientDB` between async event appending and batch file closing:

1. Multiple events queued for async append
2. Flush triggered → `fetch()` called → `finishFile()` executes
3. `finishFile()` writes closing `],"sentAt":...,"writeKey":"..."}` and sets `writer = nil`
4. Queued events still pending in async queue
5. Queued events execute, see `writer == nil`, create new writer on same file
6. New `LineStreamWriter` seeks to END of existing file (past closing bracket)
7. Events appended AFTER closing bracket = malformed JSON

**Files affected:**
- `Sources/Segment/Utilities/Storage/TransientDB.swift`
- `Sources/Segment/Utilities/Storage/Types/DirectoryStore.swift`

## Solution

Added `DispatchGroup` to track pending async append operations:

- `append()` calls `pendingAppends.enter()` before async dispatch
- `append()` calls `pendingAppends.leave()` when complete
- `fetch()` calls `pendingAppends.wait()` before closing file

This guarantees all queued async appends complete before `finishFile()` executes, preventing events from being written after the closing bracket.

### Code Changes

**TransientDB.swift:**
```swift
// Track pending async operations
private let pendingAppends = DispatchGroup()

public func append(data: RawEvent) {
    if asyncAppend {
        pendingAppends.enter()  // ← Track operation
        syncQueue.async { [weak self] in
            guard let self else {
                self?.pendingAppends.leave()
                return
            }
            store.append(data: data)
            self.pendingAppends.leave()  // ← Mark complete
        }
    }
}

public func fetch(count: Int? = nil, maxBytes: Int? = nil) -> DataResult? {
    pendingAppends.wait()  // ← Wait for pending appends
    
    var result: DataResult? = nil
    syncQueue.sync {
        result = store.fetch(count: count, maxBytes: maxBytes)
    }
    return result
}
```

## Testing

Added comprehensive test suite in `TransientDB_RaceCondition_Tests.swift`:

1. **testAsyncAppendCompletesBeforeFetch** - Verifies fix prevents race condition
2. **testSynchronousModeNoRaceCondition** - Verifies sync mode works correctly  
3. **testHighVolumeAsyncAppends** - Stress test with 100 concurrent events

All tests verify:
- Valid JSON structure (no corruption)
- Single closing bracket for batch array
- Single `sentAt` and `writeKey` fields
- All events inside batch array

## Impact

- **Severity:** Critical - silent data loss
- **Affected:** All users on default async mode (asyncAppend = true)
- **Likelihood:** Intermittent, increases with event volume and slow I/O
- **Fix:** Zero breaking changes, maintains async performance

## Checklist

- [x] Fix implemented with DispatchGroup synchronization
- [x] Comprehensive test suite added
- [x] Workaround documented for older versions
- [x] No breaking API changes
- [x] Maintains async performance (minimal blocking)
- [x] All existing tests pass

## Related

- Customer report: Batch corruption in v1.9.1
- Technical analysis: `/Users/digarcia/research/analytics-swift-batch-corruption-bug.md`

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>


BATCH_CORRUPTION_WORKAROUND.md
bsneed, line 1: don't commit this.

Sources/Segment/Utilities/Storage/TransientDB.swift
```diff
@@ -46,9 +48,14 @@ public class TransientDB {
     
     public func append(data: RawEvent) {
         if asyncAppend {
+            pendingAppends.enter()
             syncQueue.async { [weak self] in
-                guard let self else { return }
+                guard let self else {
+                    self?.pendingAppends.leave()
+                    return
```
bsneed: would this bail on doing the actual append?

bsneed (approved):

PR #421, opened 2026-03-05
Harden directory storage

bsneed:
- Add file validation before handing off to uploader.
- Check if files are finalized before appending to them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,6 @@ XCFrameworkOutput
 .idea
 .vscode
 .editorconfig
+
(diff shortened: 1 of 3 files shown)


lalmalani (approved):
PR Addresses 1 optimization to prevent a hard-to-repro race condition.
We check for file completion via a newly introduced function (BOOL) isFInished (which checks to see if there is a "sentAt" string match in the last line of file.
If isFinished returns true, we check again and start a new file with an incremented filename

PR #428, opened 2026-03-25
fix: Enable retry system for memory storage mode

MichaelGHSeg:
## Summary
Wire the retry state machine into the memory storage upload path (`flushData`), which was previously bypassed — rate limiting, backoff, max retries, and non-retryable status code drops now work in memory mode.

### SDK
- **RetryStateMachine**: Add `shouldDropBatch(statusCode:)` public method
- **HTTPClient**: Add `checkBatchUpload(batchId:)`, `shouldDropBatch(forStatusCode:)`, X-Retry-Count header support, `batchId` param for data uploads
- **SegmentDestination**: `flushData` checks retry state machine before uploading; both `flushData`/`flushFiles` drop batches on non-retryable codes; track drops via `@Atomic droppedBatchCount`
- **Analytics**: Expose `droppedBatchCount` property

### E2E CLI
- Configure `HttpConfig` with rate limiting + exponential backoff
- Use `droppedBatchCount` to detect dropped (not delivered) events
- Enable `basic,retry,settings` test suites

## Test results
All **59/59** e2e tests pass (basic 2, retry 46, settings 11)

### DIFF
--- a/Sources/Segment/Analytics.swift
+++ b/Sources/Segment/Analytics.swift
@@ -388,6 +388,14 @@ extension Analytics {
 }
 
 extension Analytics {
+    /// Returns the number of batches that were dropped (non-retryable errors or max retries exhausted).
+    public var droppedBatchCount: Int {
+        if let segmentDest = self.find(pluginType: SegmentDestination.self) {
+            return segmentDest.droppedBatchCount
+        }
+        return 0
+    }
+
     /// Determine if there are any events that have yet to be sent to Segment
     public var hasUnsentEvents: Bool {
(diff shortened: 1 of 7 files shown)


Sources/Segment/Analytics.swift
```diff
@@ -388,6 +388,14 @@ extension Analytics {
 }
 
 extension Analytics {
+    /// Returns the number of batches that were dropped (non-retryable errors or max retries exhausted).
+    public var droppedBatchCount: Int {
```
MichaelGHSeg: Eww.  Trying to do this through error handler instead.

PR #440, opened 2026-06-15
feat: honor Retry-After header on any retryable HTTP response

didiergarcia:
## Summary

- Any retryable HTTP response carrying a `Retry-After` header now triggers a pipeline-level pause, identical to 429 behavior
- Codes without `Retry-After` continue to use existing exponential backoff unchanged
- Emits a warning-level log when `Retry-After` triggers a pause on a non-429 code
- Existing 429 behavior is unchanged
- Fixed `shouldDropBatch` to correctly handle retryable codes when `rateLimitConfig.enabled = true, backoffConfig.enabled = false`

## Motivation

Per RFC 9110, `Retry-After` is a general-purpose header. Previously the SDK only respected it for 429. A 529 (or 503, 408, etc.) with `Retry-After` would be silently ignored and the SDK would compute its own (potentially shorter) backoff, increasing load on an already-overloaded server.

## Changes

- `RetryStateMachine.handleResponse()` — any retryable code with a non-nil `retryAfterSeconds` routes to `handleRateLimitResponse()` (pipeline pause, free retry) instead of `handleRetryableError()`
- `RetryStateMachine.shouldDropBatch()` — fixed to not drop batches for retryable codes when `rateLimitConfig.enabled`, regardless of `backoffConfig` state
- `HTTPClient` — warning log emitted on new pipeline pauses triggered by non-429 codes
- `RetryAfterGenericTests` — 12 new tests

## Test plan

- [x] `RetryAfterGenericTests` — all 12 tests pass
- [x] Full test suite (217 tests) passes with no regressions
- [ ] Manual: connect to test endpoint returning 529 + `Retry-After: 5`, observe pipeline pause in logs

### DIFF
--- a/Sources/Segment/Utilities/Networking/HTTPClient.swift
+++ b/Sources/Segment/Utilities/Networking/HTTPClient.swift
@@ -166,7 +166,19 @@ public class HTTPClient {
                     batchFile: batchFile,
                     currentTime: timeProvider.now()
                 )
-                retryState = stateMachine.handleResponse(state: retryState, response: responseInfo)
+                let newState = stateMachine.handleResponse(state: retryState, response: responseInfo)
+
+                // Log when Retry-After triggers a new pipeline pause on a non-429 code
+                if let retryAfterSeconds = responseInfo.retryAfterSeconds,
+                   newState.pipelineState == .rateLimited,
+                   retryState.pipelineState != .rateLimited,
+                   httpResponse.statusCode != 429 {
+                    let waitUntil = newState.waitUntilTime.map { Date(timeIntervalSince1970: $0).description } ?? "unknown"
+                    Analytics.segmentLog(message: "Retry-After (\(retryAfterSeconds)s) received on HTTP \(httpResponse.statusCode) — pausing pipeline until \(waitUntil)", kind: .warning)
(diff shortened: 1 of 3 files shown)


Sources/Segment/Utilities/Networking/HTTPClient.swift
```diff
@@ -166,7 +166,18 @@ public class HTTPClient {
                     batchFile: batchFile,
                     currentTime: timeProvider.now()
                 )
-                retryState = stateMachine.handleResponse(state: retryState, response: responseInfo)
+                let newState = stateMachine.handleResponse(state: retryState, response: responseInfo)
+
+                // Log when Retry-After triggers a new pipeline pause
```
MichaelGHSeg, line 171: I feel like this is code that could migrate into handleResponse?

didiergarcia: Good call out! The detection logic ideally belongs in handleResponse. The constraint is that RetryStateMachine is side-effect-free and has no access to a logger, so the log call has to live here in HTTPClient where the Analytics reference is in scope.

Longer term, the right fix is to change handleResponse to return a (RetryState, RetryEvent) tuple so the state machine returns what happened rather than callers figuring it out from state diffs.

MichaelGHSeg (approved):

